PR #23, opened 2025-05-02
add do command to update velocities

JohnN193:
additionally adds a canon profile to use canon instead of naked docker

currently setup to change the max vel/acc for all joints via do command, but we can probably extend this further to setting individual motors

### DIFF
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -221,12 +221,18 @@ void UR5eArm::stop(const ProtoStruct& extra) {
 ProtoStruct UR5eArm::do_command(const ProtoStruct& command) {
     ProtoStruct resp = ProtoStruct{};
     const auto vel = command.at("set_vel").get<double>();
-    if (vel != nullptr) {
-        resp.emplace("set_vel", "vel set");
-    }
-    const auto acc = command.at("set_acc").get<double>();
-    if (acc != nullptr) {
-        resp.emplace("set_acc", "acc set");
+
+    for (auto kv : command) {
+        if (kv.first == "set_vel") {
+            resp.emplace("set_vel", "vel set");
+            const double val = *kv.second.get<double>();
+            resp.emplace("set_vel_set", val);
+        }
+        if (kv.first == "set_acc") {
+            resp.emplace("set_acc", "acc set");
+            const double val = *kv.second.get<double>();
+            resp.emplace("set_acc_set", val);
+        }
     }
 
     return resp;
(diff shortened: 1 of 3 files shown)


src/ur5e_arm.cpp
```diff
@@ -267,11 +282,13 @@ void UR5eArm::move(std::vector<Eigen::VectorXd> waypoints) {
...
-    BOOST_LOG_TRIVIAL(debug) << "generating trajectory with max speed: " << speed * (180.0 / M_PI) << std::endl;
+    const double move_speed = speed.load();
+    const double move_acceleration = acceleration.load();
+    BOOST_LOG_TRIVIAL(debug) << "generating trajectory with max speed: " << move_speed * (180.0 / M_PI) << std::endl;
     Eigen::VectorXd max_acceleration(6);
     Eigen::VectorXd max_velocity(6);
-    max_acceleration << acceleration, acceleration, acceleration, acceleration, acceleration, acceleration;
-    max_velocity << speed, speed, speed, speed, speed, speed;
+    max_acceleration << move_acceleration, move_acceleration, move_acceleration, move_acceleration, move_acceleration, move_acceleration;
+    max_velocity << move_speed, move_speed, move_speed, move_speed, move_speed, move_speed;
```
JohnN193, line 291: currently we still set every join to the same max vel/acc. shouldn't be too much of a lift to separate these out if requested

src/ur5e_arm.hpp
```diff
@@ -118,6 +122,9 @@ class UR5eArm : public Arm, public Reconfigurable {
 
     // variables specified by ResourceConfig and set through reconfigure
     std::string host;
-    double speed;
-    double acceleration;
+    // double speed;
+    // double acceleration;
+
```
raybjork, line 127: remove lines

src/ur5e_arm.cpp
raybjork: I think no need to do these now we don't know how to set them well anyway

src/ur5e_arm.cpp
```diff
@@ -219,7 +219,22 @@ void UR5eArm::stop(const ProtoStruct& extra) {
...
+            resp.emplace(VEL_KEY, val);
+        }
+        if (kv.first == ACC_KEY) {
+            const double val = *kv.second.get<double>();
+            acceleration.store(val * (M_PI / 180.0));
+            resp.emplace(ACC_KEY, val);
+        }
+    }
+
+    return resp;
```
raybjork, line 237: Why return a response populated with the things you passed in?  I would expect this to return an empty struct similar to how it was before. Is there some RDK idiom with DoCommands I'm not aware of?

raybjork (approved):

PR #24, opened 2025-05-06
bump client

JohnN193:


src/ur5e_arm.cpp
```diff
@@ -86,16 +90,17 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
     }
 
     // Now the robot is ready to receive a program
-    std::unique_ptr<ToolCommSetup> tool_comm_setup;
```
JohnN193, line 89: we don't have a tool that the ur5 controls so we don't need this

src/ur5e_arm.cpp
```diff
@@ -86,16 +90,17 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
...
-                              CALIBRATION_CHECKSUM));
+    // std::unique_ptr<ToolCommSetup> tool_comm_setup;
+    urcl::UrDriverConfiguration ur_cfg = {host,
+                                          path_offset + SCRIPT_FILE,
+                                          path_offset + OUTPUT_RECIPE,
+                                          path_offset + INPUT_RECIPE,
+                                          &reportRobotProgramState,
+                                          true,  // headless mode
+                                          nullptr};
+    driver.reset(new UrDriver(ur_cfg));
```
JohnN193, line 101: switching to the config struct wasn't super necessary as they still support the older UrDriver initialization functions, but this did not appear to hurt us.

src/ur5e_arm.cpp
```diff
@@ -403,6 +408,47 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
...
+            // Release the brakes
+            if (!dashboard->commandBrakeRelease()) {
+                BOOST_LOG_TRIVIAL(info) << "yo failed release: " << status << std::endl;
+                throw std::runtime_error("couldn't release the arm brakes");
+            }
+            BOOST_LOG_TRIVIAL(info) << "yo break released: " << status << std::endl;
+
+            bool send_prog = driver->sendRobotProgram();
+            // don't know what to do if this is false yet
+            BOOST_LOG_TRIVIAL(info) << "send robot program successful: " << send_prog << std::endl;
```
JohnN193, line 441: all of these have ok bools to check, but we do not know what the behavior should be if any of these are false. we left the `throws` in the code until we have a plan

src/ur5e_arm.cpp
```diff
@@ -403,6 +408,47 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
 
 // helper function to read a data packet and send a noop message
 void UR5eArm::read_and_noop() {
+    std::string status;
+    dashboard->commandSafetyStatus(status);
+    if (status.find(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL)) == std::string::npos) {
+        estop.store(true);
```
JohnN193, line 414: prob doesn't need to be atomic, unless we want to disable other apis while the estop is triggered

src/ur5e_arm.cpp
```diff
@@ -403,6 +408,47 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
 
 // helper function to read a data packet and send a noop message
 void UR5eArm::read_and_noop() {
+    std::string status;
+    dashboard->commandSafetyStatus(status);
+    if (status.find(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL)) == std::string::npos) {
```
JohnN193, line 413: we assume any non normal safety status is an estop. this might technically be incorrect

PR #25, opened 2025-05-06
RSDK-10701 - Nicks basic cleanup

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-10701)

Done:
1. Fixes compiler warnings in our code
2. Adds logging so we can see where time is being spent (computing splines vs talking to the arm)
3. Propagate and log IO errors when talking to the arm
4. Fix cases where throwing an exception would kill the background thread
5. Write unique waypoints and trajectory files per run
6. Write the arm joint positions to a file during run 
7. Remove unused code / misleading code
8. Fix integer sign bugs in data written to csv.
9. Shut down all threads when module is shutting down
10. Make the exception safe

TODO in a later PR:
1. [Add CLI for monitoring arm positions in real time](https://github.com/viam-modules/universal-robots/pull/29/files)
2. Upgrade the viam-cpp-sdk

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.25 FATAL_ERROR)
 
+
 project(
   universal-robots
   DESCRIPTION "Viam Module to support Universal Robots arms"
@@ -20,5 +21,8 @@ FetchContent_MakeAvailable(Universal_Robots_Client_Library)
 find_package(Eigen3 3.3 REQUIRED NO_MODULE)
 find_package(viam-cpp-sdk CONFIG REQUIRED viamsdk)
 
+# add_compile_options(-Wall -Wextra -pedantic -Wno-unused-parameter -Wall -Wextra -Wpedantic -Wconversion  -Wnull-dereference -Wdouble-promotion -Wno-switch-enum -Wshadow -Wunused -Wenum-conversion -Wno-unused-function -Wuninitialized)
 add_executable(universal-robots main.cpp src/ur5e_arm.cpp trajectories/Path.cpp trajectories/Trajectory.cpp)
+add_executable(universal-robots-test test.cpp src/ur5e_arm.cpp trajectories/Path.cpp trajectories/Trajectory.cpp)
 target_link_libraries(universal-robots PRIVATE viam-cpp-sdk::viamsdk Eigen3::Eigen urcl)
+target_link_libraries(universal-robots-test PRIVATE viam-cpp-sdk::viamsdk Eigen3::Eigen urcl)
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,12 @@ format:
 build: 
 	mkdir build logs
 
-SANITIZE ?= OFF
+test: build/universal-robots
+	./build/universal-robots-test
+
 build/universal-robots: format build
 	cd build && \
-	cmake -G Ninja -DENABLE_SANITIZER=$(SANITIZE) .. && \
+	cmake -G Ninja  .. && \
 	ninja all -j 4
 
 clean: 
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -160,14 +160,15 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
     // as otherwise we will get pipeline overflows. Therefore, do this directly before starting your
     // main loop
     driver->startRTDECommunication();
-    if (!read_joint_keep_alive()) {
-        throw std::runtime_error("couldn't get joint positions");
+    while (!read_joint_keep_alive()) {
+        usleep(NOOP_DELAY);
     }
 
     // start background thread to continuously send no-ops and keep socket connection alive
     BOOST_LOG_TRIVIAL(info) << "starting background_thread";
     std::thread keep_alive_thread(&UR5eArm::keep_alive, this);
     BOOST_LOG_TRIVIAL(info) << "UR5eArm constructor end";
+    keep_alive_thread.detach();
 }
 
 // helper function to extract an attribute value from its key within a ResourceConfig
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -103,7 +103,7 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
         throw std::runtime_error("required environment variable APPDIR unset");
     }
     appdir = std::string(tmp);
-    BOOST_LOG_TRIVIAL(info) << "appdir" << appdir << "\n";
+    BOOST_LOG_TRIVIAL(info) << "appdir" << appdir;
 
     {
         std::lock_guard<std::mutex> guard{output_csv_dir_path_mu};
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -1,5 +1,7 @@
 #include "ur5e_arm.hpp"
 
+#include <cmath>
+
 // this chunk of code uses the rust FFI to handle the spatialmath calculations to turn a UR vector to a pose
 extern "C" void* quaternion_from_axis_angle(double x, double y, double z, double theta);
 extern "C" void* orientation_vector_from_quaternion(void* q);
@@ -21,6 +23,49 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
     return pose{position, orientation, theta};
 }
 
+void write_trajectory_to_file(std::string filepath,
+                              const std::vector<vector6d_t>& p_p,
+                              const std::vector<vector6d_t>& p_v,
+                              const std::vector<float>& time) {
+    bool valid = p_p.size() == p_v.size() && p_p.size() == time.size();
+    if (!valid) {
+        BOOST_LOG_TRIVIAL(info) << "write_trajectory_to_file called with invalid parameters";
+        return;
+    }
+    std::ofstream of(filepath);
+    of << "t(s),j0,j1,j2,j3,j4,j5,v0,v1,v2,v3,v4,v5\n";
+    for (size_t i = 0; i < p_p.size(); i++) {
+        of << time[i];
+        for (size_t j = 0; j < 6; j++) {
+            of << "," << p_p[i][j];
+        }
+        for (size_t j = 0; j < 6; j++) {
+            of << "," << p_v[i][j];
+        }
+        of << "\n";
+    }
+
+    of.close();
+}
+
+void write_waypoints_to_csv(std::string filepath, std::vector<Eigen::VectorXd> waypoints) {
+    unsigned i;
+    std::ofstream of(filepath);
+    for (const Eigen::VectorXd& vec : waypoints) {
+        i = 0;
+        for (const auto& n : vec) {
+            i++;
+            if (i == vec.size()) {
+                of << n;
+            } else {
+                of << n << ",";
+            }
+        }
+        of << "\n";
+    }
+    of.close();
+}
+
 // global used to track if a trajectory is in progress
 std::atomic<bool> trajectory_running(false);
 
@@ -32,7 +77,7 @@ void reportRobotProgramState(bool program_running) {
 
 // define callback function to be called by UR client library when trajectory state changes
 void reportTrajectoryState(control::TrajectoryResult state) {
-    trajectory_running = false;
+    trajectory_running.store(false);
     std::string report;
     switch (state) {
         case control::TrajectoryResult::TRAJECTORY_RESULT_SUCCESS:
@@ -49,12 +94,27 @@ void reportTrajectoryState(control::TrajectoryResult state) {
 }
 
 UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name()) {
+    BOOST_LOG_TRIVIAL(info) << "UR5eArm constructor start";
     this->reconfigure(deps, cfg);
 
     // get the APPDIR environment variable
-    path_offset = std::getenv("APPDIR");
-    if (!path_offset) {
+    auto tmp = std::getenv("APPDIR");
+    if (!tmp) {
         throw std::runtime_error("required environment variable APPDIR unset");
+    }
+    appdir = std::string(tmp);
+    BOOST_LOG_TRIVIAL(info) << "appdir" << appdir << "\n";
+
+    {
+        std::lock_guard<std::mutex> guard{output_csv_dir_path_mu};
+        if (output_csv_dir_path.empty()) {
+            tmp = std::getenv("VIAM_MODULE_DATA");
+            if (!tmp) {
+                throw std::runtime_error("required environment variable VIAM_MODULE_DATA unset");
+            }
+            output_csv_dir_path = std::string(tmp);
+            BOOST_LOG_TRIVIAL(info) << "VIAM_MODULE_DATA" << output_csv_dir_path << "\n";
+        }
     }
 
     // connect to the robot dashboard
@@ -86,11 +146,10 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
     }
 
     // Now the robot is ready to receive a program
-    // Now the robot is ready to receive a program
     urcl::UrDriverConfiguration ur_cfg = {host,
-                                          path_offset + SCRIPT_FILE,
-                                          path_offset + OUTPUT_RECIPE,
-                                          path_offset + INPUT_RECIPE,
+                                          appdir + SCRIPT_FILE,
+                                          appdir + OUTPUT_RECIPE,
+                                          appdir + INPUT_RECIPE,
                                           &reportRobotProgramState,
                                           true,  // headless mode
                                           nullptr};
@@ -101,11 +160,15 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
     // as otherwise we will get pipeline overflows. Therefore, do this directly before starting your
     // main loop
     driver->startRTDECommunication();
-    read_joint_keep_alive();
+    if (!read_joint_keep_alive()) {
+        throw std::runtime_error("couldn't get joint positions");
+    }
 
     // start background thread to continuously send no-ops and keep socket connection alive
+    BOOST_LOG_TRIVIAL(info) << "starting background_thread";
     std::thread t(&UR5eArm::keep_alive, this);
     t.detach();
+    BOOST_LOG_TRIVIAL(info) << "UR5eArm constructor end";
 }
 
 // helper function to extract an attribute value from its key within a ResourceConfig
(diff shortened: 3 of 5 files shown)


src/ur5e_arm.cpp
```diff
@@ -21,6 +23,46 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
...
+    }
+
+    std::ofstream outputFile(filepath);
+    outputFile << buffer.str();
+    outputFile.close();
+}
+
+void log_waypoints_to_csv(std::string filepath, std::vector<Eigen::VectorXd> waypoints) {
+    std::stringstream buffer;
+    for (size_t i = 0; i < waypoints.size(); i++) {
```
lia-viam, line 55: i think this can be

```cpp
for (const Eigen::VectorXd& vec : waypoints) {
    for (const auto& point : waypoints) {
        buffer << point << ", ";
    }
    buffer << "\n";
}
```

src/ur5e_arm.cpp
```diff
@@ -21,6 +23,46 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
...
+        for (size_t j = 0; j < 6; j++) {
+            buffer << waypoints[i][j] << ",";
+        }
+        buffer << std::endl;
```
lia-viam, line 59: see https://www.geeksforgeeks.org/avoid-using-stdendl/

you probably just want to be using `<< "\n"`

src/ur5e_arm.cpp
```diff
@@ -21,6 +23,46 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
...
+    }
+    std::ofstream outputFile(filepath);
```
lia-viam, line 61: `outputFile` already has an `operator<<`, so you could just get rid of `buffer` above and write to `outputFile` directly

src/ur5e_arm.cpp
```diff
@@ -135,25 +180,42 @@ void UR5eArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
...
+    for (double joint_pos_rad : joint_state) {
+        double joint_pos_deg = 180.0 / M_PI * joint_pos_rad;
+        to_ret.push_back(joint_pos_deg);
     }
     mu.unlock();
-    return std::vector<double>();
+    return to_ret;
+}
+std::chrono::milliseconds unix_now_ms() {
+    std::chrono::system_clock::time_point tp = std::chrono::system_clock::now();
```
lia-viam, line 204: consider

```cpp
namespace chrono = std::chrono;
return chrono::duration_cast<chrono::milliseconds>(system_clock::now().time_since_epoch());
```

src/ur5e_arm.cpp
```diff
@@ -135,25 +180,42 @@ void UR5eArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
...
 
 void UR5eArm::move_to_joint_positions(const std::vector<double>& positions, const ProtoStruct& extra) {
     std::vector<Eigen::VectorXd> waypoints;
     Eigen::VectorXd next_waypoint_deg = Eigen::VectorXd::Map(positions.data(), positions.size());
     Eigen::VectorXd next_waypoint_rad = next_waypoint_deg * (M_PI / 180.0);  // convert from radians to degrees
     waypoints.push_back(next_waypoint_rad);
-    move(waypoints);
+    std::chrono::milliseconds unix_time_ms = unix_now_ms();
+    log_waypoints_to_csv(path_offset + WAYPOINTS_LOG, waypoints);
+    if (!move(waypoints, unix_time_ms)) {
```
lia-viam, line 216: if you're always going to have to check the `bool` return and throw an exception if it failed it might make more sense to have `void move` which throws an exception on failure

lia-viam (commented):
mostly stylistic or hopefully helpful hints!

src/ur5e_arm.cpp
nicksanford: I'm concerned about what happens if move throws while it is holding a mutex.

Makefile
```diff
@@ -7,10 +7,12 @@ format:
 build: 
 	mkdir build logs
 
-SANITIZE ?= OFF
+test: build/universal-robots
+	./build/universal-robots-test
+
 build/universal-robots: format build
 	cd build && \
-	cmake -G Ninja -DENABLE_SANITIZER=$(SANITIZE) .. && \
```
nicksanford, line 13: this doesn't do anything

src/ur5e_arm.cpp
lia-viam: I'll include a suggestion for that below

src/ur5e_arm.cpp
```diff
@@ -416,10 +419,10 @@ bool UR5eArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisec
         }
         time.push_back(t2);
     }
-    BOOST_LOG_TRIVIAL(info) << "move: compute_trajectory end" << unix_time_ms.count() << " p.count() " << p.size() << " v " << v.size()
+    BOOST_LOG_TRIVIAL(info) << "move: compute_trajectory end " << unix_time_ms.count() << " p.count() " << p.size() << " v " << v.size()
                             << " a " << a.size() << " time " << time.size();
 
-    write_trajectory_to_file(path_offset + TRAJECTORY_LOG, p, v, a, time);
+    // write_trajectory_to_file(path_offset + TRAJECTORY_LOG, p, v, a, time);
     mu.lock();
```
lia-viam, line 445: consider the following:
```cpp
{ // note the open brace which introduces a new variable scope
    // construct a lock_guard: locks the mutex on construction and unlocks on destruction
    std::lock_guard<std::mutex> guard{mu};
    if (!send_trajectory(p, v, a, time)) {
        // no longer need to unlock the mutex before returning; lock_guard does this automatically
        BOOST_LOG_TRIVIAL(error) << "move: " << unix_time_ms.count() << " send_trajectory failed";
        return false;
    };

    bool ok = false;
    do {
        ok = read_and_noop();
    } while (trajectory_running.load() && ok);
    // once again no need to unlock, the guard goes out of scope when the brace closes below, unlocking automatically
}
```

I would replace almost all uses of manual `mu.lock()` and `mu.unlock()` with `std::lock_guard<std::mutex> lock(mu)`. In some case you may need to manually introduce a new scope with braces as I've done above, but in other cases you can just declare the guard at the beginning of the function and it will release when the function returns. 

Note that this is robust against exception throws, and also less brittle in case you later add a code path but forget to add the mutex unlock statement

src/ur5e_arm.cpp
```diff
@@ -318,21 +318,24 @@ ProtoStruct UR5eArm::do_command(const ProtoStruct& command) {
...
-void UR5eArm::keep_alive() {
+void UR5eArm::keep_alive() try {
     BOOST_LOG_TRIVIAL(info) << "keep_alive thread started";
     while (true) {
         mu.lock();
         read_and_noop();
         mu.unlock();
         usleep(NOOP_DELAY);
     }
+} catch (const std::exception& ex) {
```
lia-viam, line 348: rather than a function-level try block with a recursive call, you might just put this `try`/`catch` inside the `while` loop, and in the `catch` block log the error then call `continue`

src/ur5e_arm.cpp
nicksanford: this is awesome! Thanks for the suggestion. I've been looking for the CPP equivalent of go's:
```golang
mu.Lock()
defer mu.Unlock()
``` 
pattern

src/ur5e_arm.cpp
```diff
@@ -238,38 +337,36 @@ ProtoStruct UR5eArm::do_command(const ProtoStruct& command) {
...
-    outputFile.close();
+bool UR5eArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseconds unix_time_ms) try {
+    BOOST_LOG_TRIVIAL(info) << "move: start unix_time_ms " << unix_time_ms.count() << " waypoints size " << waypoints.size();
 
     // get current joint position and add that as starting pose to waypoints
+    BOOST_LOG_TRIVIAL(info) << "move: get_joint_positions start " << unix_time_ms.count();
     std::vector<double> curr_joint_pos = get_joint_positions(ProtoStruct{});
+    BOOST_LOG_TRIVIAL(info) << "move: get_joint_positions end" << unix_time_ms.count();
+
+    BOOST_LOG_TRIVIAL(info) << "move: compute_trajectory start " << unix_time_ms.count();
```
nicksanford, line 361: trajectory start

src/ur5e_arm.cpp
```diff
@@ -293,124 +390,166 @@ void UR5eArm::move(std::vector<Eigen::VectorXd> waypoints) {
...
+        p.push_back(vector6d_t{position[0], position[1], position[2], position[3], position[4], position[5]});
+        v.push_back(vector6d_t{velocity[0], velocity[1], velocity[2], velocity[3], velocity[4], velocity[5]});
+        float t2 = duration - (t - TIMESTEP);
+        if (std::isinf(t2)) {
+            BOOST_LOG_TRIVIAL(error) << "duration - (t - TIMESTEP) was infinite";
+            return false;
         }
+        time.push_back(t2);
     }
+    BOOST_LOG_TRIVIAL(info) << "move: compute_trajectory end " << unix_time_ms.count() << " p.count() " << p.size() << " v " << v.size()
```
nicksanford, line 441: trajectory end

Makefile
raybjork: I grabbed this makefile from an existing C++ module and didn't add in ASAN support yet

src/ur5e_arm.cpp
```diff
@@ -1,5 +1,7 @@
 #include "ur5e_arm.hpp"
 
+#include <cmath>
```
raybjork, line 3: Would someday love to learn about best practices for imports in c.  I've just been shoving everything in the hpp and importing that but I'm sure thats frowned on

src/ur5e_arm.cpp
nicksanford: done thanks!

src/ur5e_arm.cpp
nicksanford: good call, done

src/ur5e_arm.cpp
nicksanford: updated, lmk if you feel I'm using the pattern correctly

src/ur5e_arm.cpp
```diff
@@ -284,66 +386,116 @@ void UR5eArm::move(std::vector<Eigen::VectorXd> waypoints) {
...
+            write_joint_pos_deg(joint_state, pre_trajectory_state, now, 0);
+        }
+        if (!send_trajectory(p, v, time)) {
+            BOOST_LOG_TRIVIAL(error) << "move: " << unix_time_ms.count() << " send_trajectory failed";
+            return false;
+        };
+
+        std::ofstream of(arm_joint_positions_filename(viam_module_data, unix_time_ms.count()));
+
+        of << "time_ms,read_attemp,joint_0_deg,joint_1_deg,joint_2_deg,joint_3_deg,joint_4_deg,joint_5_deg\n";
```
nicksanford, line 476: typo

JohnN193 (commented):
mostly nits and a few questions

src/ur5e_arm.hpp
```diff
@@ -28,20 +29,33 @@ const std::string OUTPUT_RECIPE = "/src/control/rtde_output_recipe.txt";
...
+const std::string ARM_JOINT_POSITIONS_CSV_NAME_TEMPLATE = "/%1%_arm_joint_positions.csv";
 
 // TODO: using this is deprecated by the URCL, we could find some way around using it
 const std::string CALIBRATION_CHECKSUM = "calib_12788084448423163542";
 
 // constants for robot operation
-const int NOOP_DELAY = 100000;  // 100 milliseconds
-const double TIMESTEP = 0.2;    // seconds
+const float TIMESTEP = 0.2f;  // seconds
+const int NOOP_DELAY = 1000;  // 1 millisecond
```
JohnN193, line 41: another thing we should check, the arm has a log that states

`Setting up RTDE communication with frequency 500.0` Hz

so we should double check that setting this to 1 ms doesn't cause any issues

src/ur5e_arm.cpp
```diff
@@ -21,6 +23,49 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
     return pose{position, orientation, theta};
 }
 
+void write_trajectory_to_file(std::string filepath,
+                              const std::vector<vector6d_t>& p_p,
+                              const std::vector<vector6d_t>& p_v,
+                              const std::vector<float>& time) {
+    bool valid = p_p.size() == p_v.size() && p_p.size() == time.size();
+    if (!valid) {
+        BOOST_LOG_TRIVIAL(info) << "write_trajectory_to_file called with invalid parameters";
```
oliviamiller, line 32: should this be an error?

src/ur5e_arm.cpp
```diff
@@ -167,30 +284,42 @@ void UR5eArm::move_through_joint_positions(const std::vector<std::vector<double>
...
 pose UR5eArm::get_end_position(const ProtoStruct& extra) {
-    mu.lock();
+    std::lock_guard<std::mutex> guard{mu};
     std::unique_ptr<rtde_interface::DataPackage> data_pkg = driver->getDataPackage();
-    if (data_pkg && data_pkg->getData("actual_TCP_pose", tcp_state)) {
-        pose to_ret = ur_vector_to_pose(tcp_state);
-        mu.unlock();
-        return to_ret;
+    if (data_pkg == nullptr) {
+        BOOST_LOG_TRIVIAL(warning) << "UR5eArm::get_end_position got nullptr from driver->getDataPackage()";
```
oliviamiller, line 305: why isnt this error?

src/ur5e_arm.cpp
```diff
@@ -49,13 +94,28 @@ void reportTrajectoryState(control::TrajectoryResult state) {
...
+    if (!tmp) {
         throw std::runtime_error("required environment variable APPDIR unset");
     }
+    appdir = std::string(tmp);
+    BOOST_LOG_TRIVIAL(info) << "appdir" << appdir << "\n";
+
+    {
+        std::lock_guard<std::mutex> guard{output_csv_dir_path_mu};
+        if (output_csv_dir_path.empty()) {
+            tmp = std::getenv("VIAM_MODULE_DATA");
```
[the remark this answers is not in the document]
JohnN193: The environment variable is defined by the viam server, so we use it as the default location for data created by the module

src/ur5e_arm.cpp
```diff
@@ -292,141 +431,201 @@ void UR5eArm::move(std::vector<Eigen::VectorXd> waypoints) {
...
+        of.close();
     }
-    mu.unlock();
+
+    BOOST_LOG_TRIVIAL(info) << "move: end unix_time_ms " << unix_time_ms.count();
 }
 
 // Define the destructor
 UR5eArm::~UR5eArm() {
+    BOOST_LOG_TRIVIAL(warning) << "UR5eArm distructor called";
```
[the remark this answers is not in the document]
nicksanford: thanks done

src/ur5e_arm.cpp
```diff
@@ -292,141 +431,201 @@ void UR5eArm::move(std::vector<Eigen::VectorXd> waypoints) {
...
-            } else if (p_v.size() == time.size() && p_v[i].size() == 6) {
-                // cubic
-                driver->writeTrajectorySplinePoint(p_p[i], p_v[i], time[i]);
-            } else {
-                // linear
-                driver->writeTrajectorySplinePoint(p_p[i], time[i]);
-            }
-        }
+bool UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p, const std::vector<vector6d_t>& p_v, const std::vector<float>& time) {
+    BOOST_LOG_TRIVIAL(info) << "UR5eArm::send_trajectory start";
```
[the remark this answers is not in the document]
nicksanford: I think we want this so we can measure how much time we are spending communicating with the arm vs waiting for the arm to complete doing work.

src/ur5e_arm.cpp
```diff
@@ -292,141 +431,201 @@ void UR5eArm::move(std::vector<Eigen::VectorXd> waypoints) {
...
+    if (p_p.size() != time.size() || p_v.size() != time.size()) {
+        BOOST_LOG_TRIVIAL(error) << "UR5eArm::send_trajectory p_p.size() != time.size() || p_v.size() != time.size(): not executing";
+        return false;
+    };
+    auto point_number = static_cast<int>(p_p.size());
+    if (!driver->writeTrajectoryControlMessage(
+            urcl::control::TrajectoryControlMessage::TRAJECTORY_START, point_number, RobotReceiveTimeout::off())) {
+        BOOST_LOG_TRIVIAL(error) << "read_joint_keep_alive driver->writeTrajectoryControlMessage returned false";
```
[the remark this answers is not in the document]
nicksanford: done

src/ur5e_arm.hpp
```diff
@@ -96,32 +110,39 @@ class UR5eArm : public Arm, public Reconfigurable {
...
     void keep_alive();
-    void move(std::vector<Eigen::VectorXd> waypoints);
-    void send_trajectory(const std::vector<vector6d_t>& p_p,
-                         const std::vector<vector6d_t>& p_v,
-                         const std::vector<vector6d_t>& p_a,
-                         const std::vector<double>& time,
-                         bool use_spline_interpolation_);
-    bool read_joint_keep_alive();
+    void move(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseconds unix_time_ms);
+    bool send_trajectory(const std::vector<vector6d_t>& p_p, const std::vector<vector6d_t>& p_v, const std::vector<float>& time);
```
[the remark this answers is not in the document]
nicksanford: Can we just update the code at that time when we want to add that?

I don't see any value in a function taking parameters which are not used.

src/ur5e_arm.hpp
```diff
@@ -96,32 +110,39 @@ class UR5eArm : public Arm, public Reconfigurable {
...
     std::unique_ptr<DashboardClient> dashboard;
     vector6d_t joint_state, tcp_state;
-    std::mutex mu;
+
+    std::thread keep_alive_thread;
 
     // specified through APPDIR environment variable
-    const char* path_offset;
+    std::string appdir;
+    // specified through VIAM_MODULE_DATA environment variable
```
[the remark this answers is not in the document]
nicksanford: sorry moved

src/ur5e_arm.hpp
```diff
@@ -96,32 +110,39 @@ class UR5eArm : public Arm, public Reconfigurable {
...
     // variables specified by ResourceConfig and set through reconfigure
     std::string host;
     std::atomic<double> speed;
     std::atomic<double> acceleration;
+
+    std::mutex output_csv_dir_path_mu;
+    std::string output_csv_dir_path;
+
+    std::atomic<bool> shutdown;
+    std::atomic<bool> keep_alive_thread_alive;
```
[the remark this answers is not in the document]
nicksanford: done

src/ur5e_arm.hpp
nicksanford: that makes sense. I reduced this as when you and olivia were testing you were reporting pipe errors. I decreased this delay to try to address that. Would you like me to revert back to the previous 100 ms delay?

src/ur5e_arm.cpp
```diff
@@ -49,13 +94,28 @@ void reportTrajectoryState(control::TrajectoryResult state) {
...
+    auto tmp = std::getenv("APPDIR");
+    if (!tmp) {
         throw std::runtime_error("required environment variable APPDIR unset");
     }
+    appdir = std::string(tmp);
+    BOOST_LOG_TRIVIAL(info) << "appdir" << appdir << "\n";
+
+    {
+        std::lock_guard<std::mutex> guard{output_csv_dir_path_mu};
+        if (output_csv_dir_path.empty()) {
```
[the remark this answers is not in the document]
JohnN193: ah totally misread the code whoops

src/ur5e_arm.hpp
JohnN193: we spoke offline for this, fine to leave as is until we notice an issue/investigate if we want to change the client frequency as a whole

JohnN193 (approved):

raybjork (approved):

CMakeLists.txt
```diff
@@ -20,5 +21,8 @@ FetchContent_MakeAvailable(Universal_Robots_Client_Library)
 find_package(Eigen3 3.3 REQUIRED NO_MODULE)
 find_package(viam-cpp-sdk CONFIG REQUIRED viamsdk)
 
+# add_compile_options(-Wall -Wextra -pedantic -Wno-unused-parameter -Wall -Wextra -Wpedantic -Wconversion  -Wnull-dereference -Wdouble-promotion -Wno-switch-enum -Wshadow -Wunused -Wenum-conversion -Wno-unused-function -Wuninitialized)
```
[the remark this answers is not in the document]
nicksanford: Thanks! [Here](https://viam.atlassian.net/browse/RSDK-10721) is a ticket for this work to be done as part of a later PR

oliviamiller (approved):

PR #27, opened 2025-05-09
[RSDK-10614] Reset RTDEClient on buffer overflow or missed packet

JohnN193:
sometimes the arm will lose connection with the module. this pr will reset the connection when this behavior occurs. This includes the case when packets overflow.

### DIFF
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -415,17 +415,18 @@ bool UR5eArm::read_joint_keep_alive() {
         if (!driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP, 0, RobotReceiveTimeout::off())) {
             BOOST_LOG_TRIVIAL(error) << "read_joint_keep_alive driver->writeTrajectoryControlMessage returned false";
             return false;
-        };
+        }
     } else {
         // we received no data packet, so our comms are down. reset the comms from the driver.
-        BOOST_LOG_TRIVIAL(info) << "no packet found, resetting RTDE client connection";
+        BOOST_LOG_TRIVIAL(error) << "no packet found, resetting RTDE client connection";
         try {
             driver->resetRTDEClient(path_offset + OUTPUT_RECIPE, path_offset + INPUT_RECIPE);
-            driver->startRTDECommunication();
         } catch (const std::exception& ex) {
             BOOST_LOG_TRIVIAL(error) << "read_joint_keep_alive driver RTDEClient failed to restart: " << std::string(ex.what());
             return false;
         }
+        driver->startRTDECommunication();
+        BOOST_LOG_TRIVIAL(info) << "RTDE client connection successfully restarted";
     }
     return true;
 }


src/ur5e_arm.cpp
```diff
@@ -412,5 +412,10 @@ void UR5eArm::read_and_noop() {
 
         // send a noop to keep the connection alive
         driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP, 0, RobotReceiveTimeout::off());
+    } else {
```
raybjork, line 415: hmm it seems a little funny to be doing this resetting of communications in a function called `read_and_noop`. Thats definitely not something I would expect buried here if I were new to the codebase. Maybe its time for a better name for the function to communicate this?  The alternative to that would be erroring here and catching with this in the calling code.  That would be more verbose and I'm not sure more useful though.

JohnN193: thats fair, the concern I had with returning the error/handling this somewhere else was because read_and_noop gets used by multiple threads. would feel better about that if only keep alive was calling into there. 

maybe we could just rename this to read_from_driver? ngl I had too much fun calling the function noop that I don't remember what noop stood for. 

@nicksanford since you are also messing with these functions, not sure if you had something in mind already.

src/ur5e_arm.hpp
```diff
@@ -109,7 +109,7 @@ class UR5eArm : public Arm, public Reconfigurable {
                          const std::vector<vector6d_t>& p_a,
                          const std::vector<double>& time,
                          bool use_spline_interpolation_);
-    void read_and_noop();
+    bool read_joint_keep_alive();
```
JohnN193, line 112: updated this to return a bool to match [nick's pr](https://github.com/viam-modules/universal-robots/pull/25), but decided to skip handling the bool since the other pr already implements that behavior.

src/ur5e_arm.cpp
```diff
@@ -412,5 +412,10 @@ void UR5eArm::read_and_noop() {
...
+        // we received no data packet, so our comms are down. reset the comms from the driver.
+        BOOST_LOG_TRIVIAL(info) << "no packet found, resetting RTDE client connection" << std::endl;
```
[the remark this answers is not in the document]
JohnN193: updated

src/ur5e_arm.cpp
```diff
@@ -402,15 +402,30 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
...
+        if (!driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP, 0, RobotReceiveTimeout::off())) {
+            BOOST_LOG_TRIVIAL(error) << "read_joint_keep_alive driver->writeTrajectoryControlMessage returned false";
+            return false;
+        };
+    } else {
+        // we received no data packet, so our comms are down. reset the comms from the driver.
+        BOOST_LOG_TRIVIAL(info) << "no packet found, resetting RTDE client connection";
+        try {
+            driver->resetRTDEClient(path_offset + OUTPUT_RECIPE, path_offset + INPUT_RECIPE);
+            driver->startRTDECommunication();
```
nicksanford, line 424: I don't think this throws, at least the [header](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/2.0.0/include/ur_client_library/ur/ur_driver.h#L737) doesn't describe it as being able to throw:

src/ur5e_arm.cpp
nicksanford: I like it!

src/ur5e_arm.cpp
nicksanford: I think we can move this out of the `try`

src/ur5e_arm.cpp
```diff
@@ -402,15 +402,30 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
...
-        driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP, 0, RobotReceiveTimeout::off());
+        if (!driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP, 0, RobotReceiveTimeout::off())) {
+            BOOST_LOG_TRIVIAL(error) << "read_joint_keep_alive driver->writeTrajectoryControlMessage returned false";
+            return false;
+        };
+    } else {
+        // we received no data packet, so our comms are down. reset the comms from the driver.
+        BOOST_LOG_TRIVIAL(info) << "no packet found, resetting RTDE client connection";
+        try {
+            driver->resetRTDEClient(path_offset + OUTPUT_RECIPE, path_offset + INPUT_RECIPE);
```
nicksanford, line 423: Can we please test:
1. Start up the ur5e module with a healthy conection with the arm
2. Confirm reads & writes work
3. unplug the computer running the ur5e module from the arm
4. run reads and writes
5. confirm the module doesn't crash and that the error messages returned are clearly due to failed connectivity between the module and the arm
6. plug the ethernet back in
7. run reads & writes
8. confirm the arm works as normal

If any of the above fail (especially the error messages being unclear) I recommend tracking connectivity with the arm through an atomic and early returning from GRPC methods with a connectivity error if connectivity is not working.

nicksanford (approved):

PR #28, opened 2025-05-09
[RSDK-10672] Handle estop on the arm

JohnN193:
this pr allows the arm to recover from an estop, should one occur.

### DIFF
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -86,15 +86,15 @@ UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name())
     }
 
     // Now the robot is ready to receive a program
-   // Now the robot is ready to receive a program
+    // Now the robot is ready to receive a program
     urcl::UrDriverConfiguration ur_cfg = {host,
-        path_offset + SCRIPT_FILE,
-        path_offset + OUTPUT_RECIPE,
-        path_offset + INPUT_RECIPE,
-        &reportRobotProgramState,
-        true,  // headless mode
-        nullptr};
-driver.reset(new UrDriver(ur_cfg));
+                                          path_offset + SCRIPT_FILE,
+                                          path_offset + OUTPUT_RECIPE,
+                                          path_offset + INPUT_RECIPE,
+                                          &reportRobotProgramState,
+                                          true,  // headless mode
+                                          nullptr};
+    driver.reset(new UrDriver(ur_cfg));
     driver->registerTrajectoryDoneCallback(&reportTrajectoryState);
 
     // Once RTDE communication is started, we have to make sure to read from the interface buffer,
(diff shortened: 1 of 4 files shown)


src/ur5e_arm.cpp
```diff
@@ -403,6 +403,54 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
...
+                BOOST_LOG_TRIVIAL(info) << "arm restarted, sending control program again" << std::endl;
+
+                // send control script
+                if (!driver->sendRobotProgram()) {
+                    throw std::runtime_error("failed to resend control program");
+                }
+                BOOST_LOG_TRIVIAL(info) << "send robot program successful, restarting communication" << std::endl;
+
+                // clear any currently running trajectory.
+                trajectory_running = false;
```
JohnN193, line 441: we might want to do this when the estop is first detected. unsure of ideal behavior.

src/ur5e_arm.hpp
```diff
@@ -124,4 +124,5 @@ class UR5eArm : public Arm, public Reconfigurable {
     std::string host;
     std::atomic<double> speed;
     std::atomic<double> acceleration;
+    std::atomic<bool> estop = false;
```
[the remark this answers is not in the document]
raybjork: Are we still able to read state while Estopped?  Not sure what the best behavior would be here tbh.  I could see a case for either

src/ur5e_arm.cpp
```diff
@@ -403,6 +403,54 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
 
 // helper function to read a data packet and send a noop message
 void UR5eArm::read_and_noop() {
+    // check to see if an estop has occurred.
+    std::string status;
```
raybjork, line 407: Ok a lot more is happening in this function now.  Can we break this up into two parts?  One for managing EStops and restarting communication in fail cases and one for actually reading and nooping?  I'm open to a bunch of ideas on how to organize this tbh

src/ur5e_arm.cpp
```diff
@@ -403,6 +403,54 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
...
+    dashboard->commandSafetyStatus(status);
+
+    if (status.find(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL)) == std::string::npos) {
+        // the arm is currently estopped. save this state.
+        estop.store(true);
+    } else {
+        // the arm is in a normal state.
+        if (estop.load()) {
+            // if the arm was previously estopped, attempt to recover from the estop.
+            // We should not enter this code without the user interacting with the arm in some way(i.e. resetting the estop)
```
raybjork, line 417: How do we determine if this is the case?

src/ur5e_arm.cpp
raybjork: I think that makes sense?  Likely the trajectory has already been cancelled, but best for us to track that state internally as accurately as possible.

raybjork (changes requested):

src/ur5e_arm.cpp
JohnN193: personally I liked having all of the code live in the same function.

if peeps want me to have `read_joint_keep_alive()` call a `handle_estop` and `handle_packet`(maybe `handle_noop`?) then I'm okay with that

src/ur5e_arm.cpp
JohnN193: we know the user has switched off the estop when the `commandSafetyStatus(status)` returns `NORMAL`, and if we had already triggered a non-NORMAL status before.

src/ur5e_arm.cpp
```diff
@@ -403,6 +403,63 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
...
+                    BOOST_LOG_TRIVIAL(error)
+                        << "read_joint_keep_alive driver->sendRobotProgram() returned false when attempting to restart the arm";
+                    return false;
+                }
+                BOOST_LOG_TRIVIAL(info) << "send robot program successful, restarting communication" << std::endl;
+
+            } catch (...) {
+                BOOST_LOG_TRIVIAL(info) << "failed to restart the arm" << std::endl;
+                return false;
+            }
```
JohnN193, line 451: kept the try catch because `resetRTDEClient` can throw, but I did not confirm that the other functions in the Try do not. I also like the visual that `all of these steps must pass for us to successfully recover` that the try/catch shows

src/ur5e_arm.hpp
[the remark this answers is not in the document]
JohnN193: yes, we can still read state while estopped. unsure if you can still send commands while estopped as well

src/ur5e_arm.cpp
```diff
@@ -403,6 +403,54 @@ void UR5eArm::send_trajectory(const std::vector<vector6d_t>& p_p,
 
 // helper function to read a data packet and send a noop message
 void UR5eArm::read_and_noop() {
+    // check to see if an estop has occurred.
+    std::string status;
+    dashboard->commandSafetyStatus(status);
+
+    if (status.find(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL)) == std::string::npos) {
```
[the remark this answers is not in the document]
JohnN193: the status returns more than just `(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL))`, so this was the easiest way for us to answer "is the robot estopped". unsure if theres a better way to check if the string contains this text.

src/ur5e_arm.cpp
JohnN193: nope. when the estop is triggered we receive nothing from reportTrajectoryState, so I have to manage this myself.

src/ur5e_arm.cpp
```diff
@@ -579,6 +586,68 @@ void write_joint_pos_rad(vector6d_t js, std::ostream& of, unsigned long long uni
...
+        // the arm is currently estopped. save this state.
+        estop.store(true);
+
+        // clear any currently running trajectory.
+        trajectory_running.store(false);
+
+        // sleep longer to prevent buffer error
+        // Removing this will cause the RTDE client to move into an unrecoverable state
+        usleep(ESTOP_DELAY);
```
JohnN193, line 607: Haven’t found the reason for why we need this. All I know is the arm moves into an unrecoverable state without this sleep, getting slammed with buffer errors(which doesn’t make a ton of sense)

src/ur5e_arm.cpp
```diff
@@ -609,7 +678,7 @@ bool UR5eArm::read_joint_keep_alive(bool log) {
     }
 
     // send a noop to keep the connection alive
-    if (!driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP, 0, RobotReceiveTimeout::off())) {
+    if (!driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_CANCEL, 0, RobotReceiveTimeout::off())) {
```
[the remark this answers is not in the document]
JohnN193: 100% right we shouldn’t be, had it left over from when I was investigating something and forgot to remove

src/ur5e_arm.cpp
```diff
@@ -530,7 +537,7 @@ UR5eArm::~UR5eArm() {
     }
     BOOST_LOG_TRIVIAL(info) << "UR5eArm destructor waiting for keep_alive thread to terminate";
     while (keep_alive_thread_alive.load()) {
-        BOOST_LOG_TRIVIAL(info) << "UR5eArm destructor waiting for keep_alive thread to terminate";
+        BOOST_LOG_TRIVIAL(info) << "UR5eArm destructor still waiting for keep_alive thread to terminate";
```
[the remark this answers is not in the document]
JohnN193: Not until we bump the c++ sdk I believe

src/ur5e_arm.cpp
nicksanford: I'm ok with keeping this in one function for now.

PR #29, opened 2025-05-14
add-monitor-program

nicksanford:


Makefile
```diff
@@ -73,6 +73,14 @@ appimage-amd64: build/universal-robots
 	$(call BUILD_APPIMAGE,$(OUTPUT_NAME),$(ARCH))
 	mv ./packaging/appimages/$(OUTPUT_NAME)-*-$(ARCH).AppImage* ./packaging/appimages/deploy/
 
+
+# this embeds the text files we need to operate into a header file which can be embeded in the 
+# executable
+embed.h: ./src/control/rtde_output_recipe.txt ./src/control/rtde_input_recipe.txt ./src/control/external_control.urscript
```
raybjork, line 79: Oh word this should make a lot of the other build steps more sane.  When I go to package the AppImage I'm doing a manual copy of these files.  That might not be necessary anymore if this works how I think it does.  Would clean things up a bit

PR #31, opened 2025-05-19
RSDK-10730 - Add ursim makefile

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-10730)
1. support ur5e
2. support ur20

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: module.tar.gz ur5e-sim format test clean clean-all docker docker-build docker-amd64 docker-upload appimages docker-arm64-ci docker-amd64-ci
-default: format build/universal-robots
+default: build/universal-robots
 
 # format the source code
 format:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,22 +1,23 @@
-default: build/universal-robots
+.PHONY: module.tar.gz ur5e-sim format test clean clean-all docker docker-build docker-amd64 docker-upload appimages docker-arm64-ci docker-amd64-ci
```
nicksanford, line 1: .PHONY means `doesn't produce a file or directory with the name of the target` all these commands are phony

Makefile
```diff
@@ -1,22 +1,23 @@
...
+default: format build/universal-robots
 
 # format the source code
 format:
 	ls src/*.*pp main.cpp | xargs clang-format-15 -i --style=file
 
 build: 
-	mkdir build logs
+	mkdir build
```
nicksanford, line 9: logs directory is unused

Makefile
```diff
@@ -77,7 +78,21 @@ appimage-amd64: build/universal-robots
...
 	tar czf $@ $^ universal-robots.AppImage
+
+build/_deps/universal_robots_client_library-src/scripts/start_ursim.sh: build
+	# we need to ignore `cmake -G Ninja  ..` failing as the this project's CMakeLists.txt 
+	# assumes that the viam sdk is available as a system package (which won't be true on most
+	# development machines).
+	# However even though it fails, it will still download theuniversal-robots SDK repo which
+	# is all we need to run the sim.
+	cd build && \
+	cmake -G Ninja  ..  || true
```
JohnN193, line 92: does the` || true` make this target always return success? def okay with that just want to know what behavior, since the comment implies this target will always fail

Makefile
```diff
@@ -1,22 +1,23 @@
...
 	ls src/*.*pp main.cpp | xargs clang-format-15 -i --style=file
 
 build: 
-	mkdir build logs
+	mkdir build
 
 test: build/universal-robots
 	./build/universal-robots-test
 
-build/universal-robots: format build
```
JohnN193, line 13: can we change this back? with the current setup the appimage builds will not call `format`. its not the end of the world if I have to call format again tho

JohnN193 (approved):
just some nit Qs otherwise lgtm

Makefile
nicksanford: The target always succeeds despite the cmake command always failing

nicksanford: So yes

Makefile
nicksanford: we can't run format as format isn't always installed on the host machine, that said I'll add format to all the targets that currently depend on it

JohnN193:
Oh also do squash and merge for this pls

nicksanford:
> Oh also do squash and merge for this pls

thanks, sorry for forgetting. We should change the repo config to conform to best practices (like only allowing squash and merge).

PR #33, opened 2025-05-28
plot motion inputs versus trajectory generation inputs

nfranczak:


graph/graph.go
```diff
@@ -0,0 +1,292 @@
+package main
```
raybjork, line 1: Why is there all this go code?

graph/go.mod
```diff
@@ -0,0 +1,153 @@
+module graph.go
```
raybjork, line 1: can we call this package tools instead?

graph/graph.go
```diff
@@ -0,0 +1,292 @@
...
+func main() {
+	logger := logging.NewDebugLogger("client")
+	ctx := context.Background()
+
+	cfg, err := loadConfig("config.json")
+	if err != nil {
+		log.Fatalf("failed to load config: %v", err)
+	}
+
+	machine, err := client.New(
```
raybjork, line 66: We don't need to connect to a machine.  This go code seems like its only for converting to euclidean space right?

graph/graph_joint_inputs.py
```diff
@@ -0,0 +1,61 @@
...
+    plt.title(f'Joint {joint_idx} Position over Time')
+    plt.xlabel('Time (s)')
+    plt.ylabel('Position (rad)')
+    plt.legend()
+    plt.grid(True)
+    plt.tight_layout()
+    plt.savefig(filename)
+    plt.close()
+
+config_path = "config.json"
```
raybjork, line 21: I'm not sure what this is or why an extra config would be necessary.  You are already in the module this script should be able to know where the paths are

graph/README.md
```diff
@@ -0,0 +1,33 @@
+### Instructions
+
+In order to run we first need to create a virtual environment.
+We recommend using [uv](https://docs.astral.sh/uv/getting-started/installation/)
+Once you have  `uv` installed you need to do
+```
+uv venv -p python3.10 graph
+source graph/bin/activate
+uv pip install matplotlib numpy pandas
+```
```
randhid, line 10: .venv is idiomatic - we need to add to the .gitignore as well.
Also make a requirements file for installation of python requirements.

randhid (changes requested):
Let's spend a little more time to 
1. not require a connection to a robot - use the saved sva and trajectory files.
2. Write a more idiomatic python package.
3. Use the pandas API for less verbose plotting code.

graph/graph_joint_inputs.py
nfranczak: so suppose we running in root for some reason (this is incorrect but could happen) in this case `~/.viam/...` is not even correct.

nfranczak: you are right that this module itself is responsible for creating the files we wish to consume so perhaps it should be possible to know what the paths are, but we still need to specify the unix timestamp at the very least as well as the part-id.

randhid:
> spoke with @JohnN193 offline he has an idea to change `*trajectory.csv` from
> 
> ```
> t(s),j0,j1,j2,j3,j4,j5,v0,v1,v2,v3,v4,v5
> 0.2,-2.25382,-2.38215,-1.88255,0.540944,0.344231,-3.42488,-0,-0,0,0,0,-0
> 0.2,-2.24478,-2.40804,-1.8603,0.514576,0.334802,-3.39379,0.0635563,-0.180952,0.16296,-0.194113,-0.111703,0.22556
> 0.2,-2.24297,-2.41291,-1.85577,0.507803,0.331136,-3.38623,-0.0500574,0.125478,-0.122138,0.231094,0.118483,-0.248902
> 0.2,-2.25677,-2.38054,-1.88411,0.573643,0.342334,-3.45617,-0.0394765,0.0231278,-0.0235068,0.450666,-0.0242071,-0.433899
> ```
> 
> into
> 
> ```
> t(s),j0,j1,j2,j3,j4,j5,v0,v1,v2,v3,v4,v5
> 0.0,-2.25382,-2.38215,-1.88255,0.540944,0.344231,-3.42488,-0,-0,0,0,0,-0
> 0.2,-2.24478,-2.40804,-1.8603,0.514576,0.334802,-3.39379,0.0635563,-0.180952,0.16296,-0.194113,-0.111703,0.22556
> 0.4,-2.24297,-2.41291,-1.85577,0.507803,0.331136,-3.38623,-0.0500574,0.125478,-0.122138,0.231094,0.118483,-0.248902
> 0.6,-2.25677,-2.38054,-1.88411,0.573643,0.342334,-3.45617,-0.0394765,0.0231278,-0.0235068,0.450666,-0.0242071,-0.433899
> ```
> 
> i.e. making time not be relative.
> 
> So in terms of merging order, John's work should be merged before this PR is merged
> 
> edit: cc: @randhid @raybjork

Seems like a small orthogonal change when it does change? I wouldn't block yourself on waiting on the file format changing. None of the headers or joint inputs are changing?

raybjork:
> just to be clear, i already merged my change, and miko already updated the code to handle it

Oops sorry didn't realize. Just getting caught up from last week

nfranczak:
position_Z
position_Y
position_X
joint5_comparison
joint4_comparison
joint3_comparison
joint2_comparison
joint1_comparison
joint0_comparison

nfranczak:
position_Z
position_Y
position_X
joint5_comparison
joint4_comparison
joint3_comparison
joint2_comparison
joint1_comparison
joint0_comparison

nfranczak:
Joint 5 data looks crazy strange.
Maybe I just had a bad batch?

going to try with a different set of data.

nfranczak:
joint1_comparison

hmmm, this is worrying. Inducing a metric of time is wonky at best.

nfranczak:
actually, going back to joint5 , it's rotating by less than 1 degree. So benign imo.

graph/README.md
```diff
@@ -0,0 +1,5 @@
+### Instructions
+
+Run `make create-config`. This will create a `config.json` file which the user will need to update with two pieces of information, the part-id of the machine as well as the unix timestamp of the data they wish to graph.
```
randhid, line 3: This instruction will only work on the machine that has run the module.

raybjork (approved):
Approving because Rand has been doing a thorough job and I feel I would be redundant. Please make sure @acmorrow knows about this before you merge though

graph/graph.go
```diff
@@ -0,0 +1,351 @@
...
+			log.Printf("invalid waypoint values: %v", err)
+			continue
+		}
+		allInputs = append(allInputs, referenceframe.FloatsToInputs(vals))
+	}
+	if len(allInputs) == 0 {
+		return nil, errors.New("readWaypointCSV produced [][]referenceframe.Input of length zero")
+	}
+	return allInputs, nil
+}
```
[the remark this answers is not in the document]
nfranczak: AFAIK, there is not package that consumes a `.csv` file and returns `[]trajInput` or `[][]referenceframe.Input`

What I can do,  however, is just use `dataframe-go`

graph/Makefile
```diff
@@ -0,0 +1,11 @@
...
+	go run graph.go
+
+remove-graphs:
+	rm *.png
+
+create-config:
+	echo '{' > config.json && \
+	echo '  "trajectory_csv": "~/.viam/module-data/[part-id]/universal-robots-executable/[timestamp]_trajectory.csv",' >> config.json && \
+	echo '  "waypoints_csv": "~/.viam/module-data/[part-id]/universal-robots-executable/[timestamp]_waypoints.csv",' >> config.json && \
+	echo '}' >> config.json
```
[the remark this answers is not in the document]
nfranczak: [timestamp] would be up to the user to populate.

graph/README.md
nfranczak: not necessarily,
as long as I have waypoint and trajectory csvs this will work.

nfranczak:
spoke with @acmorrow 
we will move the files into 
```
src/viam/ur/utils/visualization
```

acmorrow:
> spoke with @acmorrow we will move the files into
> 
> ```
> src/viam/ur/utils/visualization
> ```

Thanks. There is a possibility I may end up moving them again as part of the bigger refactor as we iterate on naming and structure, but this is a good start.

acmorrow:
> > spoke with @acmorrow we will move the files into
> > ```
> > src/viam/ur/utils/visualization
> > ```
> 
> Thanks. There is a possibility I may end up moving them again as part of the bigger refactor as we iterate on naming and structure, but this is a good start.

Actually, wait. The path you said here in the comment looks right, but in the PR they are under `src/viam/utils/visualization`, so there is no `ur` directory in there, which I think we wanted.

src/viam/utils/visualization/graph.go
```diff
@@ -0,0 +1,393 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"go.viam.com/rdk/logging"
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/spatialmath"
+
+	"github.com/rocketlaunchr/dataframe-go/imports"
+	chart "github.com/wcharczuk/go-chart"
```
randhid, line 16: Why are we using a second package to plot?

nfranczak: I think this is the desired way to do plotting, linked [here](https://github.com/rocketlaunchr/dataframe-go/?tab=readme-ov-file#plotting-cross-platform) is an example from the README of `dataframe-go`

src/viam/utils/visualization/graph.go
```diff
@@ -0,0 +1,393 @@
...
+	expandedPath := path
+	if len(path) > 1 && path[:2] == "~/" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return nil, err
+		}
+		expandedPath = filepath.Join(home, path[2:])
+	}
+	return os.Open(expandedPath)
+}
```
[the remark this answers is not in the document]
randhid: I don't like how dependent this is on a string match to the second element of the filepath - please find a better way to resolve filpaths using go's filepath packages - preferable not in a little function but in the body of `realmain`

src/viam/utils/visualization/graph.go
```diff
@@ -0,0 +1,393 @@
...
+			inputValue, ok := row["j"+strconv.Itoa(j)].(float64)
+			if !ok {
+				return nil, nil, fmt.Errorf("j%d not found in row", j)
+			}
+			rowInputs = append(rowInputs, inputValue)
+		}
+		timeStep, ok := row["t(s)"].(float64)
+		if !ok {
+			return nil, nil, fmt.Errorf("t(s) not found in row")
+		}
```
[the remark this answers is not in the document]
nfranczak: > Doesn't the trajectory file always have the timestamp as recorded?

Yes, the rows of the trajectory csv apart from the header have a timestamp.

> When does this happen?

`row` has the following type `map[interface{}]interface{}`.
Hence, to get the value of `t(s)` we get it through passing in that key and  then need to assert the type as a `float64`

src/viam/utils/visualization/graph.go
```diff
@@ -0,0 +1,393 @@
...
+
+type trajPoses struct {
+	Timestamp float64
+	Pose      spatialmath.Pose
+}
+
+type config struct {
+	TrajectoryCSV string `json:"trajectory_csv"`
+	WaypointsCSV  string `json:"waypoints_csv"`
+}
```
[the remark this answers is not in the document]
nfranczak: only the config struct remains now.

src/viam/utils/visualization/graph.go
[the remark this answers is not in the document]
nfranczak: now using `"github.com/mitchellh/go-homedir"`

src/viam/utils/visualization/graph.go
```diff
@@ -0,0 +1,393 @@
...
+	var allInputs [][]referenceframe.Input
+	for i := 0; i < df.NRows(); i++ {
+		row := df.Row(i, false)
+		rowInputs := []float64{}
+		for j := 0; j < 6; j++ {
+			strVal, ok := row[j].(string)
+			if !ok {
+				return nil, nil, fmt.Errorf("key %d not found in row %d", j, i)
+			}
+			parsedVal, err := strconv.ParseFloat(strVal, 64)
```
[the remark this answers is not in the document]
nfranczak: added ✅

src/viam/utils/visualization/README.md
```diff
@@ -0,0 +1,5 @@
+### Instructions
+
+Run `make create-config`. This will create a `config.json` file which the user will need to update with two pieces of information, the part-id of the machine as well as the unix timestamp of the data they wish to graph.
```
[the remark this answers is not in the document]
nfranczak: > If the user is expected to edit the config file with the part id and the timestamp then that instruction is missing from the readme.

I will add this

> Further - you can set a module's data directory through environment variables - so perhaps we need to either have users of this code pass in the files through an argument to main or use the module's environment variable.

I am not a fan of using environment variables. My reasoning is that you set them once and then forget about them (at least I do) and then when things go awry one is left scratching their head about what when wrong.
I prefer to have all this data visible in the in the `config.json` file so that then.

Further, this module may be used on many different robots so I am unsure how "good" even using the module's environment would  be.


Lastly, suppose an individual receives csv files through slack, then they'd go into the `Downloads` directory and editing `config.json` seems easier to do than resetting an environment variable.

randhid (commented):
Mostly minor nits - found two interesting helpers from the API of rocket launch df.

Does this output well formed graphs at this stage?

src/ur/utils/visualization/graph.go
```diff
@@ -0,0 +1,311 @@
...
+	TrajectoryCSV     string `json:"trajectory_csv"`
+	WaypointsCSV      string `json:"waypoints_csv"`
+	ArmKinematicsPath string `json:"arm_kinematics_path"`
+}
+
+func loadConfig(path string) (config, error) {
+	var cfg config
+	file, err := os.Open(path)
+	if err != nil {
+		return cfg, err
```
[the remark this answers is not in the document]
randhid: Ah, sorry I thought it was a pointer 

`config{}` then, as it makes it clearer as the return value.

See pattern here: https://github.com/viamrobotics/sanding/blob/9fe0e9afb7e1ea237c70d6539b1eacde156a389b/sander.go#L508

src/ur/utils/visualization/graph.go
```diff
@@ -0,0 +1,311 @@
...
+	for i := 0; i < col.NRows(); i++ {
+		val := col.Value(i)
+		f, ok := val.(float64)
+		if !ok {
+			return nil, fmt.Errorf("column %s: value at index %d is not float64", colName, i)
+		}
+		values[i] = f
+	}
+	return values, nil
+}
```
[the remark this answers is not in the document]
nfranczak: there is a helper!

src/ur/utils/visualization/Makefile
```diff
@@ -0,0 +1,12 @@
...
+
+remove-graphs:
+	rm *.png
+
+create-example-config:
+	echo '{' > config.json && \
+	echo '  "trajectory_csv": "~/.viam/module-data/[part-id]/viam_universal-robots/[timestamp]_trajectory.csv",' >> config.json && \
+	echo '  "waypoints_csv": "~/.viam/module-data/[part-id]/viam_universal-robots/[timestamp]_waypoints.csv",' >> config.json && \
+	echo '  "arm_kinematics_path": "../../../kinematics/ur5e.json"' >> config.json && \
+	echo '}' >> config.json
```
randhid, line 12: I think you misunderstood what I said in real life.

I think I would prefer this to be an example config file rather than a command, similar to how we have a .env variable.

Right now this isn't super user-friendly if my data is copied over to Downloads, I'm having to copy paste a file path anyway. A json file that is ignored is easier to manipulate than this Makefile, which is tracked in git history and contains part ids, which someone could accidentally commit. We don't want event that committed in a public repo.

Just make an example config. This isn't helping anyone other than someone directly running this script on a machine that has run rdk locally (not through viam-agent install which captured data to `/~/root/.viam rather than `/home/.viam`, and even then you need to edit these commands.

src/ur/utils/visualization/graph_test.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/rocketlaunchr/dataframe-go"
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/test"
+)
```
randhid, line 12: Just one note on your tests - in the future please also test the unhappy path.

Test erroneous inputs, malformed path names and bad data passed in.

For example - if your trajectory planner ever produces in `NaN` or infinite joint positions, we will panic in your math function.

There is an easy fix, I believe I saw  an `NaN` sanitizer in rocket launch's df API

src/ur/utils/visualization/graph.go
```diff
@@ -191,17 +172,18 @@ func saveChartPNG(name, yLabel string, x1, y1, x2, y2 []float64) error {
...
 	}
-	suffix := []string{"X", "Y", "Z"}
 	for i := range trajData {
-		name := fmt.Sprintf("%s%d_comparison.png", prefix, i)
-		if isPos {
-			name = fmt.Sprintf("%s%s_comparison.png", prefix, suffix[i])
-		}
-		yLabel := fmt.Sprintf(yLabelFormat, i)
+		name := fmt.Sprintf("%s%s_comparison.png", prefix, labelFunc(i))
+		yLabel := fmt.Sprintf(yLabelFormat, labelFunc(i))
```
randhid, line 186: 👏🏼

src/ur/utils/visualization/Makefile
nfranczak: ah whoops

PR #36, opened 2025-06-02
Build system improvements

acmorrow:
I'll leave inline comments explaining the changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,14 +75,19 @@ RUN apt install -y libgtest-dev
 # Install Eigen
 RUN apt install -y libeigen3-dev
 
-# Install Viam C++ SDK from source
+# Install Viam C++ SDK from source. If you change the
+# version here, change it in the top level CMakeLists.txt as well.
 RUN cd /root/opt/src && \
     git clone https://github.com/viamrobotics/viam-cpp-sdk && \
     cd viam-cpp-sdk && \
     git checkout releases/v0.13.2 && \
-    mkdir build && \
-    cd build && \
-    cmake  -DCMAKE_BUILD_TYPE=RelWithDebInfo -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON -DCMAKE_INSTALL_PREFIX=/usr/local .. -G Ninja  && \
-    ninja all -j 4 && \
-	ninja install -j 4 && \
+    cmake -S . -B build \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON \
+        -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON \
+        -DVIAMCPPSDK_BUILD_EXAMPLES=OFF \
+        -DVIAMCPPSDK_BUILD_TESTS=OFF \
+        -G Ninja && \
+    cmake --build build --target all -- -j4 && \
+    cmake --install build --prefix /usr/local && \
     rm -rf /root/opt/src/viam-cpp-sdk
--- a/Makefile
+++ b/Makefile
@@ -5,18 +5,18 @@ default: build/universal-robots
 format:
 	ls src/*.*pp main.cpp | xargs clang-format-15 -i --style=file
 
-build: 
+build:
 	mkdir build
 
 test: build/universal-robots
 	./build/universal-robots-test
 
 build/universal-robots: build
-	cd build && \
-	cmake -G Ninja  .. && \
-	ninja all -j 4
+	cmake -S . -B build -G Ninja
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -75,14 +75,19 @@ RUN apt install -y libgtest-dev
...
 RUN cd /root/opt/src && \
     git clone https://github.com/viamrobotics/viam-cpp-sdk && \
     cd viam-cpp-sdk && \
     git checkout releases/v0.13.2 && \
-    mkdir build && \
-    cd build && \
-    cmake  -DCMAKE_BUILD_TYPE=RelWithDebInfo -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON -DCMAKE_INSTALL_PREFIX=/usr/local .. -G Ninja  && \
-    ninja all -j 4 && \
-	ninja install -j 4 && \
+    cmake -S . -B build \
```
acmorrow, line 84: Cmake can happily make the `build` directory for you

Dockerfile
```diff
@@ -75,14 +75,19 @@ RUN apt install -y libgtest-dev
...
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON \
+        -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON \
+        -DVIAMCPPSDK_BUILD_EXAMPLES=OFF \
+        -DVIAMCPPSDK_BUILD_TESTS=OFF \
+        -G Ninja && \
+    cmake --build build --target all -- -j4 && \
+    cmake --install build --prefix /usr/local && \
```
acmorrow, line 92: It is usually better to drive builds by using cmake directly, rather than dropping down into the generated build system, unless you need to.

Note the addition of the `cmake --install` step, which is the right way to do installs, because it can rewrite things like rpath appropriately as the install is placed in different places.

Makefile
```diff
@@ -5,18 +5,18 @@ default: build/universal-robots
...
 test: build/universal-robots
 	./build/universal-robots-test
 
 build/universal-robots: build
-	cd build && \
-	cmake -G Ninja  .. && \
-	ninja all -j 4
+	cmake -S . -B build -G Ninja
+	cmake --build build --target all -- -j4
+	cmake --install build --prefix build/install/usr
```
acmorrow, line 17: As above, CMake can act as the driver for the configure and build steps, and use `cmake --install` to place it in `build/install/usr`. This `prefix` is actually a bit of a lie, since the files are actually going to get staged into the appimage under `/usr/`. That's OK actually, since `/usr/lib` and friends are on the default library search path.

I'm somewhat interested in taking a closer look at this, because what I actually want is:

`cmake --install build --prefix /usr --destdir build/install`. In other words, put an installation designed for placement in `/usr` into `build/install`.

main.cpp
```diff
@@ -12,6 +13,8 @@
 using namespace viam::sdk;
 
 int main(int argc, char** argv) {
+    const Instance instance;
```
acmorrow, line 16: Oops. I missed this in #35. I thought I'd tested adequately but I guess I failed to rebuild the docker image correctly.

packaging/appimages/universal-robots-aarch64.yml
```diff
@@ -3,8 +3,7 @@ version: 1
 script:
  - rm -rf AppDir || true
  - mkdir -p $TARGET_APPDIR/usr/bin $TARGET_APPDIR/usr/lib $TARGET_APPDIR/src/
- - cp ../../build/universal-robots $TARGET_APPDIR/usr/bin/
- - cp ../../build/_deps/universal_robots_client_library-build/liburcl.so $TARGET_APPDIR/usr/lib/
+ - cp -r ../../build/install/usr/* $TARGET_APPDIR/usr
```
acmorrow, line 6: Now we don't need to pick out the individual libraries, or reach into subdirectories of `build` like `_deps` to get our targets: they are all staged into `build/install` for us. So just bulk copy that tree into the `/usr` tree of the app image.

packaging/appimages/universal-robots-x86_64.yml
```diff
@@ -3,8 +3,7 @@ version: 1
 script:
  - rm -rf AppDir || true
  - mkdir -p $TARGET_APPDIR/usr/bin $TARGET_APPDIR/usr/lib $TARGET_APPDIR/src/
- - cp ../../build/universal-robots $TARGET_APPDIR/usr/bin/
- - cp ../../build/_deps/universal_robots_client_library-build/liburcl.so $TARGET_APPDIR/usr/lib/
+ - cp -r ../../build/install/usr/* $TARGET_APPDIR/usr
```
acmorrow, line 6: As above.

raybjork:
I think I'm unqualified as a code reviewer for this so I am going to remove myself. If the build/deploy processes are unaffected in quality it is good with me

PR #38, opened 2025-06-04
Move arm driver state into owned struct instance

acmorrow:
This is a purely mechanical change, but it will allow us to entirely drop (or not have) an internal state for a driver instance.

Cases where this is interesting/desirable:
- After `stop` has completed, all internal state can be dropped by calling `current_state.reset()`.
- `reconfigure` can now start by effecting a `stop`, resulting in a fully empty state, then try and build a new state.
- After a failed `reconfigure`, the driver instance will not have any state.

FYI this was mostly done by Claude, please review accordingly.

### DIFF
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -97,6 +97,7 @@ void reportTrajectoryState(control::TrajectoryResult state) {
 
 UR5eArm::UR5eArm(Dependencies deps, const ResourceConfig& cfg) : Arm(cfg.name()) {
     BOOST_LOG_TRIVIAL(info) << "UR5eArm constructor start";
+    current_state_ = std::make_unique<state_>();
     this->reconfigure(deps, cfg);
 
     // get the APPDIR environment variable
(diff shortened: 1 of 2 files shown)


nicksanford (approved):

nicksanford:
> This is a purely mechanical change, but it will allow us to entirely drop (or not have) an internal state for a driver instance.
> 
> Cases where this is interesting/desirable:
> 
> * After `stop` has completed, all internal state can be dropped by calling `current_state.reset()`.
> * `reconfigure` can now start by effecting a `stop`, resulting in a fully empty state, then try and build a new state.
> * After a failed `reconfigure`, the driver instance will not have any state.
> 
> FYI this was mostly done by Claude, please review accordingly.

Good with this change. FYI I don't think we want to throw away all state when `stop` is called. I think we just want to stop physical execution.

randhid:
> After stop has completed, all internal state can be dropped by calling current_state.reset().

We certainly don't want this at all. Stop should not clear any state - it should only command the bytes to stop the arm's motors from moving. You meant this for a future pr yes?

> reconfigure can now start by effecting a stop, resulting in a fully empty state, then try and build a new state.

FYI, reconfigure is going away at some point as per NetCode's eng plan - so I suggest just making it an empty function and relying on the constructor and destructor to initialise and store the state of the driver. We'd have to make sure that the C++ SDK supports always rebuild behaviour with the SDK team. I can talk more about what that is.

Also - yes we want to stop the arm, but Stop API should not be the arbitrator of the internal state that gets cleared; it should do what it is describing - Stop moving the physical arm.  I would never want it to clear an e-stop state , for example. I would want a user to go over to the arm and clear the state from the physical controlelr, or have a well-defined DoCommand to clear that state. 

> After a failed reconfigure, the driver instance will not have any state.
I'd expect the resource manager to remove the resource from our resource graph after a failed reconfigure, and the destructor to get called, which would be the appropriate place to clear the state IMO - is that not happening? Might be a C++ SDK bug.

randhid:
Changes themselves in this pr are okay with me, adding @JohnN193 for his thoughts!

PR #45, opened 2025-06-13
[RSDK-10875] header cleanup

mattjperez:
mechanical changes to minimize header footprint/linkage

compiles, tested on simulator.

### DIFF
--- a/src/ur5e_arm.hpp
+++ b/src/ur5e_arm.hpp
@@ -86,14 +86,7 @@ class UR5eArm final : public Arm, public Reconfigurable {
    private:
     struct state_;
 
-    enum class UrDriverStatus : int8_t  // Only available on 3.10/5.4
-    {
-        NORMAL = 1,
-        ESTOPPED = 2,
-        READ_FAILURE = 3,
-        DASHBOARD_FAILURE = 4
-    };
-
+    enum class UrDriverStatus : int8_t;  // Only available on 3.10/5.4
     static std::string status_to_string(UrDriverStatus status);
 
     void keep_alive();
--- a/src/ur5e_arm.cpp
+++ b/src/ur5e_arm.cpp
@@ -14,6 +14,15 @@
 
 #include "../trajectories/Trajectory.h"
 
+// this chunk of code uses the rust FFI to handle the spatialmath calculations to turn a UR vector to a pose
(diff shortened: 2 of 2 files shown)


src/ur5e_arm.hpp
```diff
@@ -2,46 +2,16 @@
...
-// constants for robot operation
-constexpr float TIMESTEP = 0.2F;     // seconds
-constexpr int NOOP_DELAY = 2000;     // 2 millisecond/500 Hz
-constexpr int ESTOP_DELAY = 100000;  // 100 millisecond/10 Hz
-
-// do_command keys
-constexpr char VEL_KEY[] = "set_vel";
-constexpr char ACC_KEY[] = "set_acc";
-
-void reportRobotProgramState(bool program_running);
```
JohnN193, line 44: do we not want to declare this here still?

JohnN193 (commented):
looks like these are all changes that I don't really have context as for why we should do them. lmk if you want me to review further

src/ur5e_arm.hpp
```diff
@@ -2,46 +2,16 @@
 
 #include <ur_client_library/control/trajectory_point_interface.h>
 #include <ur_client_library/types.h>
```
acmorrow, line 4: Do you still need both includes? Or can you get away with just `types.h` now?

src/ur5e_arm.hpp
```diff
@@ -113,47 +82,29 @@ class UR5eArm final : public Arm, public Reconfigurable {
...
+    struct state_;
+
     enum class UrDriverStatus : int8_t  // Only available on 3.10/5.4
     {
         NORMAL = 1,
         ESTOPPED = 2,
         READ_FAILURE = 3,
         DASHBOARD_FAILURE = 4
     };
+
```
acmorrow, line 96: It may be possible to forward declare UrDriverStatus, but I'm not 100%. Let's try it together and see.

acmorrow (changes requested):
This is looking good. A few small suggestions, let's do them together at your desk when you are free.

src/ur5e_arm.hpp
mattjperez: needed both

src/ur5e_arm.hpp
mattjperez: was able to move to definition to cpp

PR #46, opened 2025-06-13
Add kinematics files for ur20

randhid:
Added the kinematics files into the repo in  `src/kinematics` - we prefer the `ur20.json` file over the `ur20.urdf`, since that allows us to generate capsules for the geometries associated with each link of the arm.

### DIFF
--- a/src/kinematics/ur20.json
+++ b/src/kinematics/ur20.json
@@ -1,0 +1,261 @@
+{
+    "name": "UR20",
+    "kinematic_param_type": "SVA",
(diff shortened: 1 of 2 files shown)


randhid:
> LGTM as long as you just want approval to add these files as they are. If you want me to actually review the contents, let me know and I will do so, but I don't expect that's terribly useful.

Add files as they are - I want to get them out of my local fork and into main for you to use. These are the kinematics that I've been using on my local fork.

PR #48, opened 2025-06-16
[RSDK-10978] Add support for UR20

acmorrow:
- Rename the `ur5e_arm.*` files to `ur_arm.*` since they are no longer specific to UR5e.
- Rename the `UR5eArm` class to `URArm`, since it is no longer specific to UR5e.
- Create a static function `URArm::create_model_registrations` which returns all relevant model registrations, and return one each for UR20 and UR5e, then use this function in main, which no longer needs to know anything about specific models.
- Feed the `Model` object back into the `URArm` constructor so instances know which arm model they represent.
- Switch on the current `Model` to select the correct kinematics file when asked at runtime.


acmorrow:
@randhid - I've done a quick check against the simulator and I was able to bring up the simulator in both UR5e and UR20 mode and adjust the model in my machine config and control the arm.

src/ur_arm.cpp
```diff
@@ -153,16 +153,35 @@ struct UR5eArm::state_ {
...
+
+    const auto arm = API::get<Arm>();
+    const auto registration_factory = [&](const Model& model) {
+        return std::make_shared<ModelRegistration>(arm, model, std::bind(arm_factory, model, _1, _2));
+    };
+
+    const auto family = ModelFamily{"viam", "universal-robots"};
+    return {
+        registration_factory({family, "ur5e"}),
+        registration_factory({family, "ur20"}),
```
JohnN193, line 179: we should go ahead and add this to the meta.json. we can prob wait to update the readme before we release

src/ur_arm.cpp
```diff
@@ -358,60 +377,73 @@ void UR5eArm::move_through_joint_positions(const std::vector<std::vector<double>
...
-bool UR5eArm::is_moving() {
+bool URArm::is_moving() {
     return current_state_->trajectory_running.load();
 }
 
-UR5eArm::KinematicsData UR5eArm::get_kinematics(const ProtoStruct&) {
+URArm::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
+    // The `Model` class absurdly lacks accessors
+    const std::string model_string = [&] {
+        const auto family = ModelFamily{"viam", "universal-robots"};
```
JohnN193, line 401: whats the rationale for not making this a const on URArm? we only need it here and registration currently?

JohnN193 (approved):
Tested on the ur5e and ur20. both appeared in viz tab correctly(eyeballed the end effector pose) and ran correctly with both arm and motion apis. 

could still test with the actual sanding setup, since I derived all of my motion plans using joint positions rather than feedback from the camera. 

only actual ask is updating the meta.json & maybe the readme with the new model, but those can technically wait until we are closer to release

src/ur_arm.cpp
acmorrow: Done

src/ur_arm.cpp
acmorrow: It could be. It'd be nicer if ModelFamily had a constexpr ctor, but it doesn't. But I can make a static function on URArm to return that ModelFamily, and then I can also add a static `model` function. That'd get rid of some of the duplication fairly nicely.

acmorrow: Done

acmorrow (commented):
I've made the requested updates, PTAL.

PR #50, opened 2025-06-17
[RSDK-10805] cleanup

mattjperez:
small cleanup stuff, including:
- rename constants from CAPS to kNaming, i.e. `MY_CONST` -> `kMyConst` (google c++ style guide)
- move consts that are used only once into their respective functions
- use `std::chrono::milliseconds` type instead of int (better type representation)
- use `std::this_thread::sleep_for` instead of `usleep` (better portability to non-UNIX systems)

compiles, ~testing on simulator~, works on simulator

### DIFF
--- a/src/ur_arm.cpp
+++ b/src/ur_arm.cpp
@@ -555,23 +555,23 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
             throw std::runtime_error("trajectory.getDuration() was infinite");
         }
         float t = 0.0;
-        constexpr float kTimestep = 0.2F;  // seconds
+        constexpr float k_timestep = 0.2F;  // seconds
         while (t < duration) {
             Eigen::VectorXd position = trajectory.getPosition(t);
             Eigen::VectorXd velocity = trajectory.getVelocity(t);
             p.push_back(vector6d_t{position[0], position[1], position[2], position[3], position[4], position[5]});
             v.push_back(vector6d_t{velocity[0], velocity[1], velocity[2], velocity[3], velocity[4], velocity[5]});
-            time.push_back(kTimestep);
-            t += kTimestep;
+            time.push_back(k_timestep);
+            t += k_timestep;
         }
 
         Eigen::VectorXd position = trajectory.getPosition(duration);
         Eigen::VectorXd velocity = trajectory.getVelocity(duration);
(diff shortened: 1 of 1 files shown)


src/ur_arm.cpp
```diff
@@ -25,20 +26,12 @@ extern "C" void free_quaternion_memory(void* q);
...
-constexpr char SVA_FILE_TEMPLATE[] = "%1%/src/kinematics/%2%.json";
-constexpr char SCRIPT_FILE[] = "/src/control/external_control.urscript";
-constexpr char OUTPUT_RECIPE[] = "/src/control/rtde_output_recipe.txt";
-constexpr char INPUT_RECIPE[] = "/src/control/rtde_input_recipe.txt";
-
-// locations of log files that will be written
-constexpr char TRAJECTORY_CSV_NAME_TEMPLATE[] = "/%1%_trajectory.csv";
-constexpr char WAYPOINTS_CSV_NAME_TEMPLATE[] = "/%1%_waypoints.csv";
-constexpr char ARM_JOINT_POSITIONS_CSV_NAME_TEMPLATE[] = "/%1%_arm_joint_positions.csv";
+constexpr char kOutputRecipe[] = "/src/control/rtde_output_recipe.txt";
```
acmorrow, line 29: This repo is inconsistent, but I think I'd prefer `k_output_recipe` over `kOutputRecipe`, etc.

src/ur_arm.cpp
```diff
@@ -25,20 +26,12 @@ extern "C" void free_quaternion_memory(void* q);
...
+constexpr char kInputRecipe[] = "/src/control/rtde_input_recipe.txt";
 
 // constants for robot operation
-constexpr float TIMESTEP = 0.2F;     // seconds
-constexpr int NOOP_DELAY = 2000;     // 2 millisecond/500 Hz
-constexpr int ESTOP_DELAY = 100000;  // 100 millisecond/10 Hz
+constexpr std::chrono::milliseconds kNoopDelay = std::chrono::milliseconds(2000);     // 2 millisecond/500 Hz
```
acmorrow, line 33: `auto` works here and is entirely unambiguous:
```
constexpr auto k_noop_delay = std::chrono::milliseconds(2000);
```

acmorrow (changes requested):

acmorrow (approved):
LGTM

PR #52, opened 2025-06-17
Fix workflow for 0.2.0-rc2

acmorrow:


JohnN193 (approved):

.github/workflows/build.yml
```diff
@@ -2,8 +2,8 @@
 on:
     push:
```
JohnN193, line 3: should this be removed?

PR #54, opened 2025-06-18
[RSDK-11002] reorg workspace

mattjperez:
this PR:
1. moves `main.cpp`, `test.cpp`, and `ur_arm.[c,h]pp` into a new `src/viam/ur/` subdirectory
2. moves the `trajectories` submodule into a new `src/third_party/` subdirectory

Compiles, test passes, runs on physical arm without issue


CMakeLists.txt
```diff
@@ -88,8 +88,8 @@ add_library(trajectories)
 set_target_properties(trajectories PROPERTIES SYSTEM TRUE)
 target_sources(trajectories
   PRIVATE
-    trajectories/Trajectory.cpp
-    trajectories/Path.cpp
+    src/third_party/trajectories/Trajectory.cpp
+    src/third_party/trajectories/Path.cpp
```
mattjperez, line 92: I think the better solution will be creating a `CMakeLists.txt` in `src/third_party` that can manage all those, though not familiar enough with CMake yet to make these optimizations

src/viam/ur/ur_arm.cpp
```diff
@@ -14,7 +14,7 @@
 #include <viam/sdk/registry/registry.hpp>
 #include <viam/sdk/resource/resource.hpp>
 
-#include "../trajectories/Trajectory.h"
+#include "../../third_party/trajectories/Trajectory.h"
```
mattjperez, line 17: I think we can make this better with the build system, but not sure how yet.

src/viam/ur/ur_arm.hpp
```diff
@@ -8,7 +8,7 @@
 #include <viam/sdk/registry/registry.hpp>
 #include <viam/sdk/resource/reconfigurable.hpp>
 
-#include "../trajectories/Path.h"
+#include "../../third_party/trajectories/Path.h"
```
mattjperez, line 11: same here

mattjperez:
@nfranczak just tagging you since I know you're adding a bunch of files along the same directory structure in #33

acmorrow:
@mattjperez - Don't forget to look in the workflow directory for references to these files under their old paths.

acmorrow:
@mattjperez - Also the top-level makefile

acmorrow (approved):

PR #56, opened 2025-06-18
[RSDK-11020] Restore magic add-repository call

acmorrow:
Apparently, the duplicated line was needed: https://unix.stackexchange.com/questions/776941/apt-add-repository-needs-to-run-twice-on-debian-12


acmorrow:
@mattjperez or @JohnN193 - I broke the docker build, will merge on whoever approves first.

JohnN193 (approved):
love that

PR #58, opened 2025-06-20
add support for UR3e

raybjork:
It would be great to support another member of the UR family :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,9 @@ build/_deps/universal_robots_client_library-src/scripts/start_ursim.sh: build
 	cd build && \
 	cmake -G Ninja  ..  || true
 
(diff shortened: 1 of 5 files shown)


acmorrow:
Mechanical changes to define the new model look good. Please let me know when you are happy with the kinematics file and I'll approve it for you to merge.

raybjork:
Screenshot 2025-06-24 at 6 57 08 PM
With the latest commit on this branch I now have a kinematics file that does a reasonable job of approximating the kinematics.  It would be really valuable to run this in sim though, since we don't actually have access to one of these arms at the moment.  @acmorrow how difficult do you expect this will be?

acmorrow:
I was able to run it against the simulator no problem and send it motion commands which looked reasonably executed. I think you are free to merge this if you like.

acmorrow (approved):

PR #59, opened 2025-06-20
[RSDK-10878] Treat reconfiguration for UR arms as always rebuild

acmorrow:
This more or less gives the UR arm module the "always rebuild" behavior for reconfigure. I'll leave some comments on the PR as well.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -2,6 +2,7 @@
 
 #include <chrono>
 #include <cmath>
+#include <stdexcept>
 #include <thread>
 
 #include <boost/format.hpp>
@@ -101,7 +102,7 @@ auto make_scope_guard(Callable&& cleanup) {
        private:
         std::function<void()> cleanup_;
     };
-    return guard{std::move(cleanup)};
+    return guard{std::forward<Callable>(cleanup)};
 }
 
 }  // namespace
@@ -344,14 +345,14 @@ void URArm::startup_(const Dependencies&, const ResourceConfig& cfg) {
     current_state_->driver.reset(new UrDriver(ur_cfg));
 
     // define callback function to be called by UR client library when trajectory state changes
-    current_state_->driver->registerTrajectoryDoneCallback(std::bind(&URArm::trajectory_done_cb, this, std::placeholders::_1));
+    current_state_->driver->registerTrajectoryDoneCallback(std::bind(&URArm::trajectory_done_cb_, this, std::placeholders::_1));
 
     // Once RTDE communication is started, we have to make sure to read from the interface buffer,
     // as otherwise we will get pipeline overflows. Therefore, do this directly before starting your
     // main loop
     current_state_->driver->startRTDECommunication();
     int retry_count = 100;
-    while (read_joint_keep_alive(false) != UrDriverStatus::NORMAL) {
+    while (read_joint_keep_alive_(false) != UrDriverStatus::NORMAL) {
         if (retry_count <= 0) {
             throw std::runtime_error("couldn't get joint positions; unable to establish communication with the arm");
         }
@@ -361,13 +362,21 @@ void URArm::startup_(const Dependencies&, const ResourceConfig& cfg) {
 
     // start background thread to continuously send no-ops and keep socket connection alive
     VIAM_SDK_LOG(info) << "starting background_thread";
-    current_state_->keep_alive_thread = std::thread(&URArm::keep_alive, this);
+    current_state_->keep_alive_thread = std::thread(&URArm::keep_alive_, this);
 
     VIAM_SDK_LOG(info) << "URArm startup complete";
     failure_handler.deactivate();
 }
 
-void URArm::trajectory_done_cb(const control::TrajectoryResult state) {
+void URArm::check_configured_() {
+    if (!current_state_) {
+        std::ostringstream buffer;
+        buffer << "Arm is not currently configured; reconfiguration likely failed";
+        throw std::runtime_error(buffer.str());
+    }
+}
+
+void URArm::trajectory_done_cb_(const control::TrajectoryResult state) {
     std::string report;
     switch (state) {
         case control::TrajectoryResult::TRAJECTORY_RESULT_SUCCESS:
@@ -395,11 +404,12 @@ void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
 }
 
 std::vector<double> URArm::get_joint_positions(const ProtoStruct&) {
+    check_configured_();
     if (current_state_->local_disconnect.load()) {
         throw std::runtime_error("arm is currently in local mode");
     }
     const std::lock_guard<std::mutex> guard{current_state_->mu};
-    if (read_joint_keep_alive(true) == UrDriverStatus::READ_FAILURE) {
+    if (read_joint_keep_alive_(true) == UrDriverStatus::READ_FAILURE) {
         throw std::runtime_error("failed to read from arm");
     };
     std::vector<double> to_ret;
@@ -443,6 +453,8 @@ std::string URArm::get_output_csv_dir_path() {
 }
 
 void URArm::move_to_joint_positions(const std::vector<double>& positions, const ProtoStruct&) {
+    check_configured_();
+
     if (current_state_->local_disconnect.load()) {
         throw std::runtime_error("arm is currently in local mode");
     }
@@ -460,12 +472,14 @@ void URArm::move_to_joint_positions(const std::vector<double>& positions, const 
     write_waypoints_to_csv(filename, waypoints);
 
     // move will throw if an error occurs
-    move(waypoints, unix_time_ms);
+    move_(waypoints, unix_time_ms);
 }
 
 void URArm::move_through_joint_positions(const std::vector<std::vector<double>>& positions,
                                          const MoveOptions&,
                                          const viam::sdk::ProtoStruct&) {
+    check_configured_();
+
     if (current_state_->local_disconnect.load()) {
         throw std::runtime_error("arm is currently in local mode");
     }
@@ -486,11 +500,13 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
         write_waypoints_to_csv(filename, waypoints);
 
         // move will throw if an error occurs
-        move(waypoints, unix_time_ms);
+        move_(waypoints, unix_time_ms);
     }
 }
 
 pose URArm::get_end_position(const ProtoStruct&) {
+    check_configured_();
+
     if (current_state_->local_disconnect.load()) {
         throw std::runtime_error("arm is currently in local mode");
     }
@@ -508,10 +524,13 @@ pose URArm::get_end_position(const ProtoStruct&) {
 }
 
 bool URArm::is_moving() {
+    check_configured_();
     return current_state_->trajectory_status.load() == TrajectoryStatus::k_running;
 }
 
 URArm::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
+    check_configured_();
+
     // The `Model` class absurdly lacks accessors
     const std::string model_string = [&] {
         if (model_ == model("ur5e")) {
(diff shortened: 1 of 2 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -68,6 +68,24 @@ T find_config_attribute(const ResourceConfig& cfg, const std::string& attribute)
 
 enum class TrajectoryStatus { k_running = 1, k_cancelled = 2, k_stopped = 3 };
 
+template <typename Callable>
```
acmorrow, line 71: Newer versions of boost have this, but not the one in our docker image, so I've rolled my own kinda "meh" one, but good enough.

This is basically how you write "finally" in C++.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -185,9 +203,35 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
 }
 
 URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) : Arm(cfg.name()), model_(std::move(model)) {
-    VIAM_SDK_LOG(info) << "URArm constructor start (model: " << model_.to_string() << ")";
+    VIAM_SDK_LOG(info) << "URArm constructor called (model: " << model_.to_string() << ")";
+    startup_(deps, cfg);
```
acmorrow, line 207: All work previously done in the constructor or reconfigure is now done in `startup_`.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -185,9 +203,35 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
...
+    }
+
+    VIAM_SDK_LOG(info) << "URArm starting up";
     current_state_ = std::make_unique<state_>();
-    this->reconfigure(deps, cfg);
+
+    // If we fail to make it through the startup sequence, execute the shutdown code. The
+    // shutdown code must be prepared to be called from any intermediate state that this
+    // function may have constructed due to partial execution.
+    auto failure_handler = make_scope_guard([&] {
```
acmorrow, line 221: If we fail during `startup_` we need to `shutdown_`. This scope guard ensures that happens, unless we deactivate the guard, which we do at the end of `startup_`.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -679,19 +721,43 @@ std::string URArm::status_to_string(UrDriverStatus status) {
...
+            }
+            if (current_state_->keep_alive_thread.joinable()) {
+                VIAM_SDK_LOG(info) << "URArm shutdown waiting for keep_alive thread to terminate";
+                current_state_->keep_alive_thread.join();
+                VIAM_SDK_LOG(info) << "keep_alive thread terminated";
+            }
+            VIAM_SDK_LOG(info) << "URArm shutdown complete";
+        }
+    } catch (...) {
+        const auto unconditional_abort = make_scope_guard([] { std::abort(); });
```
acmorrow, line 752: This is open for discussion. As written, if we can't shut down one arm, we will terminate the process which will kill all instances using this module. I wrote it this way because it wasn't clear to me what other recourse we have, but I'm interested in alternatives.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -290,16 +336,12 @@ void URArm::trajectory_done_cb(const control::TrajectoryResult state) {
...
-    current_state_->speed.store(find_config_attribute<double>(cfg, "speed_degs_per_sec") * (M_PI / 180.0));
-    current_state_->acceleration.store(find_config_attribute<double>(cfg, "acceleration_degs_per_sec2") * (M_PI / 180.0));
-    try {
-        const std::lock_guard<std::mutex> guard{current_state_->output_csv_dir_path_mu};
-        current_state_->output_csv_dir_path = find_config_attribute<std::string>(cfg, "csv_output_path");
-    } catch (...) {  // NOLINT: TODO: What should actually happen if the attribute is missing?
-    }
+void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
+    VIAM_SDK_LOG(warn) << "Reconfigure called: shutting down current state";
+    shutdown_();
```
JohnN193, line 341: [q] do we need to worry about the arm apis getting called while this/startup is in progress?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -679,19 +721,43 @@ std::string URArm::status_to_string(UrDriverStatus status) {
...
+    VIAM_SDK_LOG(warn) << "URArm destructor called, shutting down";
+    shutdown_();
+    VIAM_SDK_LOG(warn) << "URArm destroyed";
+}
+
+void URArm::shutdown_() noexcept {
+    try {
+        VIAM_SDK_LOG(warn) << "URArm shutdown called";
+        if (current_state_) {
+            const auto destroy_state = make_scope_guard([&] { current_state_.reset(); });
```
acmorrow, line 733: This ensures that we go back to a state where we can call `startup_` (current_state_ is empty) as a result of calling `shutdown_`.

acmorrow:
Note that this doesn't really address the core concurrency concerns I have with the current implementation. But it does offer hope that if you did, say, a reconfiguration where you changed the target host for an arm, that the reconfigured instance would actually start talking to the new device.

src/viam/ur/module/ur_arm.cpp
acmorrow: We do, yes. However, I was hoping to stack that work above this, because I think a bigger overhaul of locking is required.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -185,9 +203,35 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
...
 URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) : Arm(cfg.name()), model_(std::move(model)) {
-    VIAM_SDK_LOG(info) << "URArm constructor start (model: " << model_.to_string() << ")";
+    VIAM_SDK_LOG(info) << "URArm constructor called (model: " << model_.to_string() << ")";
+    startup_(deps, cfg);
+}
+
+void URArm::startup_(const Dependencies&, const ResourceConfig& cfg) {
+    if (current_state_) {
+        throw std::logic_error("URArm::startup_ was called for an already running instance");
+    }
```
randhid, line 213: [q] When are we expecting to startup an already running instance? Or is this a guard against this function being called from another one of the Arm's APIs.

I would expect this to only be called in the constructor for this driver.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -185,9 +203,35 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
...
+    });
+
+    // extract relevant attributes from config
+    current_state_->host = find_config_attribute<std::string>(cfg, "host");
+    current_state_->speed.store(find_config_attribute<double>(cfg, "speed_degs_per_sec") * (M_PI / 180.0));
+    current_state_->acceleration.store(find_config_attribute<double>(cfg, "acceleration_degs_per_sec2") * (M_PI / 180.0));
+    try {
+        const std::lock_guard<std::mutex> guard{current_state_->output_csv_dir_path_mu};
+        current_state_->output_csv_dir_path = find_config_attribute<std::string>(cfg, "csv_output_path");
+    } catch (...) {  // NOLINT: TODO: What should actually happen if the attribute is missing?
```
JohnN193, line 233: below this we set `output_csv_dir_path` to whatever is in `VIAM_MODULE_DATA` if output_csv_dir_path is empty. maybe we can just move that logic into the catch?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -185,9 +203,35 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
...
+    // function may have constructed due to partial execution.
+    auto failure_handler = make_scope_guard([&] {
+        VIAM_SDK_LOG(warn) << "URArm startup failed - shutting down";
+        shutdown_();
+    });
+
+    // extract relevant attributes from config
+    current_state_->host = find_config_attribute<std::string>(cfg, "host");
+    current_state_->speed.store(find_config_attribute<double>(cfg, "speed_degs_per_sec") * (M_PI / 180.0));
+    current_state_->acceleration.store(find_config_attribute<double>(cfg, "acceleration_degs_per_sec2") * (M_PI / 180.0));
```
randhid, line 229: I want to confirm that missing these attributes ensures that the arm fails to build, which is the current behaviour.

Note - speed and acceleration could be set to safe defaults in the future  if that results in less friction in configuration.

src/viam/ur/module/ur_arm.cpp
randhid, line 233: Either make a default in viam/module/data or warn - this is more for us to debug the trajectory generation code and make this driver stable.

src/viam/ur/module/ur_arm.cpp
JohnN193: kinda spooky but I think this is fine. tbh not sure what would have to happen for disconnect to throw, and we aren't that worried about the behavior of multiple arms yet

randhid: Hmm. 

Questions:
- Could we ever shut down the entire module process while an actuating command is in progress on one of the arms? That seems ... not great.
- More generally - if we can't safely use the universal robots RTDE APIs to stop and terminate a connection to the arm safely, that's not something we want. 
- I think this design is probably safe if we only have one robot resource per instance of  viam-server - which I don't really like.

JohnN193 (approved):
LGTM

src/viam/ur/module/ur_arm.cpp
acmorrow: This gets called again if you reconfigure, in order to start up with the newly provided configuration. The check here ensures that `startup_` is never called with an active configuration already in place.

src/viam/ur/module/ur_arm.cpp
acmorrow: > I want to confirm that missing these attributes ensures that the arm fails to build, which is the current behaviour.

Yes, it will. The `find_config_attribute` throws on error. If this is reached from the constructor, creation of the arm object will fail. If it is reached from `reconfigure`, then reconfiguration will fail and the arm instance will not be in a configured state.

> 
> Note - speed and acceleration could be set to safe defaults in the future if that results in less friction in configuration.

I don't think it is really causing any friction to require them.

src/viam/ur/module/ur_arm.cpp
acmorrow: This code didn't change in this review. It just moved. I'm trying to make this review as close to simple code motion as I can.

acmorrow: See above: this isn't new code, just moved, so I don't really want to change it now.

src/viam/ur/module/ur_arm.cpp
acmorrow: Note that the logic here is a superset of what was in the destructor before, and if you throw from a destructor that isn't tagged `noexcept(false)` then the C++ runtime behavior is to automatically call `std::terminate`. So if you removed an arm from your configuration (which would result in the destructor being called), and that failed with an exception, then the whole module would terminate. So in a sense what is happening here isn't at all new - but it is made explicit and logged.

It is true that the scope of this limitation/behavior is now broader, because now you can reach this as part of reconfiguring an arm instance, not just removing it.

I also agree that we should try to do better here. For instance, trying to reconfigure an arm from which we are disconnected may currently be crashy. But before, it didn't even actually do anything, and in fact it could result in your reconnecting to the wrong arm!

JohnN193:
yeah i was wondering about that [here](https://github.com/viam-modules/universal-robots/pull/59#discussion_r2159607515), but I figured with the planned refactors it was okay not to worry about for now.

acmorrow:
@JohnN193 and @randhid - How do you want to proceed with this review? I can either:
- Merge it as is, even though we know the locking isn't right, as an intermediate step. That'd imply a commitment to fix the locking for the next RC.
- Hold these changes back, close this review, and I can build the locking above this. That'd make the next review have this work too, but I could leave it as a separate commit and you could review just the commits above it.

It sort of depends on whether you want to leave the head state of the repo as fully correct (such as it may be) or are OK with having work-in-progress state on `main` between RCs / releases.

PR #60, opened 2025-06-21
[RSDK-11004] validate arm model

mattjperez:
This PR checks that the arm model the module is configured for is the same as the connected arm.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -190,16 +190,16 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
     // check model type is valid, map to ur_client data type
     // https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
     const std::string configured_model_type = [&] {
-        const std::string cmodel = model_.to_string();
-        if (cmodel == "viam:universal-robots:ur3e") {
+        const std::string cfg_model = model_.to_string();
+        if (cfg_model == "viam:universal-robots:ur3e") {
             return "UR3";
-        } else if (cmodel == "viam:universal-robots:ur5e") {
+        } else if (cfg_model == "viam:universal-robots:ur5e") {
             return "UR5";
-        } else if (cmodel == "viam:universal-robots:ur20e") {
+        } else if (cfg_model == "viam:universal-robots:ur20e") {
             return "UR20";
         } else {
             std::ostringstream buffer;
-            buffer << "unsupported model type: `" << cmodel << "`";
+            buffer << "unsupported model type: `" << cfg_model << "`";
             throw std::invalid_argument(buffer.str());
         }
     }();
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -186,6 +186,24 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
 
 URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) : Arm(cfg.name()), model_(std::move(model)) {
     VIAM_SDK_LOG(info) << "URArm constructor start (model: " << model_.to_string() << ")";
+
+    // check model type is valid, map to ur_client data type
+    // https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
+    const std::string configured_model_type =
+        [&] {
+            if (model_ == "viam:universal-robots:ur5e") {
+                return "UR5";
+            } else if (model_ == "viam:universal-robots:ur20e") {
+                return "UR20";
+            } else if (model_ == "viam:universal-robots:ur3e") {
+                return "UR3"
+            } else {
+                std::ostringstream buffer;
+                buffer << "invalid model type: `" + model_ + "`";
+                throw std::invalid_argument(buffer.str());
+            }
+        }
+
     current_state_ = std::make_unique<state_>();
     this->reconfigure(deps, cfg);
 
@@ -216,15 +234,14 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
     }
 
     // verify that the connected arm is the same model as the configured module
-    std::string connected_model_type{};
-    if (!current_state_->dashboard->commandGetRobotModel(connected_model_type)) {
+    std::string actual_model_type{};
+    if (!current_state_->dashboard->commandGetRobotModel(actual_model_type)) {
         throw std::runtime_error("failed to get model info of connected arm");
     }
 
-    std::string configured_model_type = model_type();
-    if (configured_model_type != connected_model_type) {
+    if (configured_model_type != actual_model_type) {
         std::ostringstream buffer;
-        buffer << "configured model type `" + configured_model_type + "` does not match connected arm `" + connected_model_type + "`";
+        buffer << "configured model type `" + configured_model_type + "` does not match connected arm `" + actual_model_type + "`";
         throw std::runtime_error(buffer.str());
     }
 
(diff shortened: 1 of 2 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -49,6 +49,17 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
     return pose{position, orientation, theta};
 }
 
+// https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
+std::string to_robot_type_string(const std::string& model) {
```
mattjperez, line 53: if we have even more models we might want to change this to a map of some kind. doing so now feels a bit premature

src/viam/ur/module/ur_arm.cpp
```diff
@@ -215,6 +226,20 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
...
+        if (!current_state_->dashboard->commandGetRobotModel(conn_model_type)) {
+            throw std::runtime_error("failed to get model info of connected arm");
+        }
+
+        std::string cfg_model_type = to_robot_type_string(model_.to_string());
+        if (cfg_model_type != conn_model_type) {
+            const auto err = "configured model type `" + cfg_model_type + "` does not match connected arm `" + conn_model_type + "`";
+            throw std::runtime_error(err);
+        }
+    }
```
mattjperez, line 241: scoped so we don't carry the strings longer than necessary

src/viam/ur/module/ur_arm.cpp
```diff
@@ -215,6 +226,20 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
...
+    // verify that the connected arm is the same model as the configured module
+    {
+        std::string conn_model_type{};
+        if (!current_state_->dashboard->commandGetRobotModel(conn_model_type)) {
+            throw std::runtime_error("failed to get model info of connected arm");
+        }
+
+        std::string cfg_model_type = to_robot_type_string(model_.to_string());
+        if (cfg_model_type != conn_model_type) {
+            const auto err = "configured model type `" + cfg_model_type + "` does not match connected arm `" + conn_model_type + "`";
```
mattjperez, line 238: verified that this error shows up on the component card's error tab.

tried with separate error/debug logs, but it wasn't as easy to find in the noise of the other logs

src/viam/ur/module/ur_arm.cpp
```diff
@@ -49,6 +49,17 @@ pose ur_vector_to_pose(urcl::vector6d_t vec) {
     return pose{position, orientation, theta};
 }
 
+// https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
+std::string to_robot_type_string(const std::string& model) {
+    if (model == "viam:universal-robots:ur5e") {
+        return "UR5";
+    }
+    if (model == "viam:universal-robots:ur20e") {
```
mattjperez, line 57: there's probably a better way for us to carry the `model_` info, but when combined with the limited return types offered by `commandGetRobotModel` this seemed the most straightforward approach

src/viam/ur/module/ur_arm.cpp
acmorrow: Good to consider the resource implications, but here I wouldn't bother. It is just the constructor, which won't ever be on a fast path, and we can assume basically no memory pressure compared to the size of these strings. And I'd expect the compiler's lifetime analysis to figure it out anyway.

src/viam/ur/module/ur_arm.cpp
acmorrow: A better way to build up a string is with an `ostringstream`. See https://github.com/viam-modules/viam-mlmodelservice-triton/blob/3026fade112f2406d5b82fab3a570db9b8a8648e/src/viam_mlmodelservice_triton_impl.cpp#L749-L753 for instance

src/viam/ur/module/ur_arm.cpp
```diff
@@ -508,6 +522,18 @@ void URArm::keep_alive() {
     VIAM_SDK_LOG(info) << "keep_alive thread terminating";
 }
 
+// https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
+std::string URArm::model_type() {
+    std::string model = model_.to_string();
+    if (model == "viam:universal-robots:ur5e") {
```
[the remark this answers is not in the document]
acmorrow: - You don't even need a real function for this. Have a look at what I did to solve the same problem in `get_kinematics` with the lambda.
- Can it be DRY'd with the logic in `get_kinematics` without hassle? If so, then it could warrant a named function. But I think the returned strings are slightly different.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -508,6 +522,18 @@ void URArm::keep_alive() {
...
+        return "UR5";
+    }
+    if (model == "viam:universal-robots:ur20e") {
+        return "UR20";
+    }
+    return "UNRECOGNIZED_MODEL_TYPE:" + model;
```
acmorrow, line 534: - Add a check to the constructor that the `Model` argument is a supported one (I should have done this), and throw std::invalid_argument if it is not.
- Throw std::logic_error from this function if somehow we managed to get into this without a valid model, rather than making a non-meaningful return value.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -215,6 +215,20 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
...
     }
 
+    // verify that the connected arm is the same model as the configured module
+    {
+        std::string conn_model_type{};
+        if (!current_state_->dashboard->commandGetRobotModel(conn_model_type)) {
+            throw std::runtime_error("failed to get model info of connected arm");
+        }
+
+        std::string cfg_model_type = model_type();
```
acmorrow, line 225: - const, but see below
- I'd call this `configured_model_type` - no need to shorten it. Autocomplete means you never need to type it anyway.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -215,6 +215,20 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
...
+        if (cfg_model_type != conn_model_type) {
```
acmorrow, line 226: C++17 allows you to put an initialization inside a conditional:
```
if (const auto configured_model_type = model_type(); configured_model_type != actual_model_type) {
   ...
}
```

src/viam/ur/module/ur_arm.cpp
```diff
@@ -215,6 +215,20 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
         throw std::runtime_error("couldn't connect to dashboard");
     }
 
+    // verify that the connected arm is the same model as the configured module
+    {
+        std::string conn_model_type{};
```
acmorrow, line 220: I'd call this `actual_model_type` or `connected_model_type`

acmorrow (changes requested):

acmorrow:
Just FYI because you and @raybjork are likely to have some collisions here: https://github.com/viam-modules/universal-robots/pull/58

src/viam/ur/module/ur_arm.cpp
mattjperez: moved the logic for getting the model type to the top of the constructor to throw the invalid argument error

src/viam/ur/module/ur_arm.cpp
```diff
@@ -186,6 +186,24 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
 
 URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) : Arm(cfg.name()), model_(std::move(model)) {
     VIAM_SDK_LOG(info) << "URArm constructor start (model: " << model_.to_string() << ")";
+
+    // check model type is valid, map to ur_client data type
+    // https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
+    const std::string configured_model_type = [&] {
+        const std::string cfg_model = model_.to_string();
+        if (cfg_model == "viam:universal-robots:ur3e") {
```
mattjperez, line 194: in anticipation of ray's PR

acmorrow (approved):
LGTM

PR #61, opened 2025-06-23
[RSDK-11035] fix clang-tidy,clang-check errors

mattjperez:


src/viam/ur/module/ur_arm.cpp
```diff
@@ -84,7 +91,7 @@ T find_config_attribute(const ResourceConfig& cfg, const std::string& attribute)
     return *val;
 }
 
-enum class TrajectoryStatus { k_running = 1, k_cancelled = 2, k_stopped = 3 };
+enum class TrajectoryStatus : std::uint8_t { k_running = 1, k_cancelled = 2, k_stopped = 3 };
```
mattjperez, line 94: we talked about ignoring it, but I like having an example of this in the code for future reference. would rather have this than more NOLINTS

Makefile
```diff
@@ -11,6 +11,10 @@ build:
 test: build/universal-robots
 	./build/universal-robots-test
 
+tidy: build/universal-robots
```
mattjperez, line 14: was torn between `make tidy` and `make bless`. seemed a bit unnecessary to make them make commands

src/viam/ur/module/ur_arm.cpp
```diff
@@ -620,36 +620,36 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
...
-        constexpr float k_timestep = 0.2F;  // seconds
+        double t = 0.0;
+        constexpr double k_timestep = 0.2;  // seconds
         while (t < duration) {
             Eigen::VectorXd position = trajectory.getPosition(t);
             Eigen::VectorXd velocity = trajectory.getVelocity(t);
             p.push_back(vector6d_t{position[0], position[1], position[2], position[3], position[4], position[5]});
             v.push_back(vector6d_t{velocity[0], velocity[1], velocity[2], velocity[3], velocity[4], velocity[5]});
-            time.push_back(k_timestep);
+            time.push_back((float)k_timestep);
```
mattjperez, line 634: we convert to float here to not bloat the log files

src/viam/ur/module/test.cpp
```diff
@@ -77,17 +77,17 @@ BOOST_AUTO_TEST_CASE(test_write_trajectory_to_file) {
 }
 
 BOOST_AUTO_TEST_CASE(test_waypoints_filename) {
-    auto x = waypoints_filename("/home/user", 1747161493357);
+    auto x = waypoints_filename("/home/user", std::chrono::milliseconds(1747161493357));
```
acmorrow, line 80: Cool trick: use the literals: https://en.cppreference.com/w/cpp/chrono/operator%2522%2522ms.html

```
using namespace std::chrono_literals;
auto x = waypoints_filename("...", 1747161493357us)
```

Makefile
acmorrow, line 14: Let's split these into different targets, called `run-clang-tidy` and `run-clang-check`.

src/viam/ur/module/ur_arm.cpp
acmorrow, line 94: I actually disagree with the complaint here. We don't care about the size of this. Let the compiler pick. Can you NOLINT it? I might care in headers so I don't want to disable the check completely.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -414,8 +410,9 @@ void URArm::move_to_joint_positions(const std::vector<double>& positions, const
     const Eigen::VectorXd next_waypoint_deg = Eigen::VectorXd::Map(positions.data(), boost::numeric_cast<Eigen::Index>(positions.size()));
     const Eigen::VectorXd next_waypoint_rad = next_waypoint_deg * (M_PI / 180.0);  // convert from radians to degrees
     waypoints.push_back(next_waypoint_rad);
-    const std::chrono::milliseconds unix_time_ms = unix_now_ms();
-    auto filename = waypoints_filename(get_output_csv_dir_path(), unix_time_ms.count());
+    const auto unix_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
```
acmorrow, line 413: I don't think you need `ms` on the variable, since it is clearly milliseconds per the duration cast, here and elsewhere.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -496,8 +496,8 @@ URArm::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
     const std::streamsize fileSize = sva_file.tellg();
     sva_file.seekg(0, std::ios::beg);
 
-    // Create a buffer to hold the file contents
-    std::vector<unsigned char> kinematics_bytes(fileSize);
+    // Create a buffer to hold the file contents, int long fits in unsigned long
+    std::vector<unsigned char> kinematics_bytes((unsigned long)fileSize);
```
acmorrow, line 500: Don't use C-style casts. Use `static_cast` for that. Also, I think it is worth thinking a bit more about what this warning was telling us. If `streamsize` can be negative, I don't think you want to cast it to positive and then allocate that many bytes.

src/viam/ur/module/ur_arm.cpp
acmorrow, line 634: - Don't use C style casts.
- Since it looks like `time` really holds floats, this needs to check for overflow.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -724,6 +725,7 @@ std::string URArm::status_to_string(UrDriverStatus status) {
 }
 
 // Define the destructor
+// NOLINT(bugprone-exception-escape) 'may throw' warning
```
acmorrow, line 728: This will go away with my reconfigure changes, but fine to leave in for now.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -414,8 +410,9 @@ void URArm::move_to_joint_positions(const std::vector<double>& positions, const
     const Eigen::VectorXd next_waypoint_deg = Eigen::VectorXd::Map(positions.data(), boost::numeric_cast<Eigen::Index>(positions.size()));
     const Eigen::VectorXd next_waypoint_rad = next_waypoint_deg * (M_PI / 180.0);  // convert from radians to degrees
     waypoints.push_back(next_waypoint_rad);
-    const std::chrono::milliseconds unix_time_ms = unix_now_ms();
-    auto filename = waypoints_filename(get_output_csv_dir_path(), unix_time_ms.count());
+    const auto unix_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
+
+    auto filename = waypoints_filename(get_output_csv_dir_path(), unix_time_ms);
```
acmorrow, line 415: I was hoping that you would just be able to do:
```
auto filename = waypoints_filename(get_output_csv_dir_path(), std::chrono::system_clock::now().time_since_epoch()));
```

If that doesn't work and the duration cast is needed, then I'd say put back the `unix_now_ms` function so you don't need to write so much. Sorry, I may have sent you on the wrong path on that one.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -620,36 +620,36 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
...
         Eigen::VectorXd velocity = trajectory.getVelocity(duration);
         p.push_back(vector6d_t{position[0], position[1], position[2], position[3], position[4], position[5]});
         v.push_back(vector6d_t{velocity[0], velocity[1], velocity[2], velocity[3], velocity[4], velocity[5]});
-        const float t2 = duration - (t - k_timestep);
+        const double t2 = duration - (t - k_timestep);
         if (std::isinf(t2)) {
             throw std::runtime_error("duration - (t - k_timestep) was infinite");
         }
-        time.push_back(t2);
+        time.push_back((float)t2);
```
acmorrow, line 646: As above.

acmorrow (changes requested):

Makefile
acmorrow: Also these targets should be phony: we want to run them even if `build/universal-robots` is up to date.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -35,6 +35,12 @@ constexpr char k_input_recipe[] = "/src/control/rtde_input_recipe.txt";
 constexpr auto k_noop_delay = std::chrono::milliseconds(2);     // 2 millisecond, 500 Hz
 constexpr auto k_estop_delay = std::chrono::milliseconds(100);  // 100 millisecond, 10 Hz
 
+// define callback function to be called by UR client library when program state changes
+void reportRobotProgramState(bool program_running) {
+    // Print the text in green so we see it better
+    VIAM_SDK_LOG(info) << "\033[1;32mUR program running: " << std::boolalpha << program_running << "\033[0m";
```
acmorrow, line 41: Would you please file a ticket to 1) remove the use of color logging in this repo, and 2) ensure that we aren't permanently altering the stream state? If the boost logging stuff isn't doing its own stream state protection, then we need to.

src/viam/ur/module/ur_arm.cpp
acmorrow: I actually think it is more than that: I think the UR driver actually takes this as float:
- https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/ur_driver.h#L511

src/viam/ur/module/ur_arm.cpp
mattjperez: [RSDK-11048](https://viam.atlassian.net/browse/RSDK-11048?atlOrigin=eyJpIjoiMjliOTNlNWE3MzhiNDQ2NDgzODQxZDU3ZTc3YzViMzUiLCJwIjoiaiJ9)

Makefile
```diff
@@ -11,6 +11,12 @@ build:
 test: build/universal-robots
 	./build/universal-robots-test
 
+run-clang-tidy:
+	clang-tidy-19 --config-file ./.clang-tidy -p build ./src/viam/ur/module/*.[h,c]pp
```
acmorrow, line 15: Per discussion, just pass the .cpp files to clang-tidy (and I think clang-check as well), since they are the only thing known to the compilation database. Also, add a `--header-filter` that selects only the headers under `viam/ur/module` since those are the only ones we are interested in.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -497,7 +501,7 @@ URArm::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
     sva_file.seekg(0, std::ios::beg);
 
     // Create a buffer to hold the file contents
-    std::vector<unsigned char> kinematics_bytes(fileSize);
+    std::vector<unsigned char> kinematics_bytes(static_cast<unsigned long>(fileSize));
```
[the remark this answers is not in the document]
acmorrow: Covered on slack.

Here is a better way:
```
    std::ifstream sva_file(sva_file_path, std::ios::binary);
    if (!sva_file) {
        throw std::runtime_error(str(boost::format("unable to open kinematics file '%1'") % sva_file_path));
    }

    // Read the entire file into a vector without computing size ahead of time
    std::vector<char> temp_bytes(std::istreambuf_iterator<char>(sva_file), {});
    if (sva_file.bad()) {
        throw std::runtime_error(str(boost::format("error reading kinematics file '%1'") % sva_file_path));
    }

    // Convert to unsigned char vector
    return KinematicsDataSVA({temp_bytes.begin(), temp_bytes.end()});
```

The copy is unfortunate but unavoidable given the somewhat odd choice of `unsigned char` for the return vector.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -84,7 +97,7 @@ T find_config_attribute(const ResourceConfig& cfg, const std::string& attribute)
     return *val;
 }
 
-enum class TrajectoryStatus { k_running = 1, k_cancelled = 2, k_stopped = 3 };
+enum class TrajectoryStatus { k_running = 1, k_cancelled = 2, k_stopped = 3 };  // NOLINT
```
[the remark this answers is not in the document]
acmorrow: You can also `NOLINTNEXTLINE` which might be a little cleaner.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -724,7 +728,7 @@ std::string URArm::status_to_string(UrDriverStatus status) {
 }
 
 // Define the destructor
-URArm::~URArm() {
+URArm::~URArm() {  // NOLINT
```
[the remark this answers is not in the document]
acmorrow: Also ditto `NOLINTNEXTLINE`

acmorrow (commented):
Change the `NOLINT`s to `NOLINTNEXTLINE(reason-for-lint)` and I'll approve this.

acmorrow (approved):
LGTM. If `make run-clang-tidy run-clang-check` are passing in the docker container, I suggest merging this ASAP so it doesn't go stale.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -788,6 +788,7 @@ std::string URArm::status_to_string_(UrDriverStatus status) {
 }
 
 // Define the destructor
+// NOLINTNEXTLINE(bugprone-exception-escape)
```
acmorrow, line 791: FYI you may not need this after merging from main.

PR #62, opened 2025-06-25
disable cancelled errors

JohnN193:
disables the throw that occurs when a trajectory gets cancelled, as certain move calls would trigger this error reliably

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -755,7 +755,7 @@ void URArm::move_(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseco
         of.close();
         VIAM_SDK_LOG(info) << "move: end unix_time " << unix_time.count();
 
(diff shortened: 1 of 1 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -755,9 +755,11 @@ void URArm::move_(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseco
         of.close();
         VIAM_SDK_LOG(info) << "move: end unix_time " << unix_time.count();
 
-        if (current_state_->trajectory_status.load() == TrajectoryStatus::k_cancelled) {
-            throw std::runtime_error("arm's current trajectory cancelled by code");
-        }
+        // TODO: renable this once zero velocity errors are resolved
```
acmorrow, line 758: nit: Usually I put the ticket number in the todo like `TODO(RSDK-11063)`. Meanwhile, please also add a note in RSDK-11063 containing a reminder to re-enable this logic.

PR #63, opened 2025-06-26
[RSDK-10979] remove duplicate joint positions and handle invalid durations

JohnN193:
removes duplicate joint positions from the positions provided by the user, which helps the trajectory repo generate valid trajectories. 

additionally returns early from a move call if the arm is already at the goal position and no other waypoints were specified, as well as erroring if we detect an invalid trajectory via a bad `duration`
ticket: https://viam.atlassian.net/browse/RSDK-10979

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -36,6 +36,8 @@ constexpr char k_input_recipe[] = "/src/control/rtde_input_recipe.txt";
 constexpr auto k_noop_delay = std::chrono::milliseconds(2);     // 2 millisecond, 500 Hz
 constexpr auto k_estop_delay = std::chrono::milliseconds(100);  // 100 millisecond, 10 Hz
 
+constexpr double k_waypoint_equivalancy_epsilon_rad = 1e-4;
+
 // define callback function to be called by UR client library when program state changes
 void reportRobotProgramState(bool program_running) {
     // Print the text in green so we see it better
@@ -498,7 +500,7 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
             const Eigen::VectorXd next_waypoint_deg =
                 Eigen::VectorXd::Map(position.data(), boost::numeric_cast<Eigen::Index>(position.size()));
             const Eigen::VectorXd next_waypoint_rad = next_waypoint_deg * (M_PI / 180.0);  // convert from radians to degrees
-            if ((!waypoints.empty()) && (next_waypoint_rad.isApprox(waypoints.back(), 0.0001))) {
+            if ((!waypoints.empty()) && (next_waypoint_rad.isApprox(waypoints.back(), k_waypoint_equivalancy_epsilon_rad))) {
                 continue;
             }
             waypoints.push_back(next_waypoint_rad);
@@ -640,9 +642,10 @@ void URArm::move_(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseco
     const Eigen::VectorXd curr_waypoint_deg =
         Eigen::VectorXd::Map(curr_joint_pos.data(), boost::numeric_cast<Eigen::Index>(curr_joint_pos.size()));
     const Eigen::VectorXd curr_waypoint_rad = curr_waypoint_deg * (M_PI / 180.0);
-    if (!curr_waypoint_rad.isApprox(waypoints.front(), 0.0001)) {
+    if (!curr_waypoint_rad.isApprox(waypoints.front(), k_waypoint_equivalancy_epsilon_rad)) {
         waypoints.insert(waypoints.begin(), curr_waypoint_rad);
-    } else if (waypoints.size() < 2) {  // this tells us if we are already at the goal
+    } 
+    if (waypoints.size() == 1) {  // this tells us if we are already at the goal
         VIAM_SDK_LOG(debug) << "arm is already at the desired joint positions";
         return;
     }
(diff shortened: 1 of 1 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -270,9 +270,10 @@ URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) :
     current_state_->keep_alive_thread = std::thread(&URArm::keep_alive, this);
     VIAM_SDK_LOG(info) << "URArm constructor end";
 }
-
+int rando_counter = 0;
```
acmorrow: testing?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -383,10 +384,14 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
     // TODO: use options
     if (!positions.empty()) {
         std::vector<Eigen::VectorXd> waypoints;
+
         for (auto position : positions) {
             const Eigen::VectorXd next_waypoint_deg =
                 Eigen::VectorXd::Map(position.data(), boost::numeric_cast<Eigen::Index>(position.size()));
             const Eigen::VectorXd next_waypoint_rad = next_waypoint_deg * (M_PI / 180.0);  // convert from radians to degrees
+            if ((!waypoints.empty()) && (next_waypoint_rad.isApprox(waypoints.back()))) {
```
acmorrow: What's the epsilon here for `isApprox`? I think we should put it in a named constant, with units. I think these are radians so `k_waypoint_equivalancy_epsilon_rad` or something.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -520,7 +528,12 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
     const Eigen::VectorXd curr_waypoint_deg =
         Eigen::VectorXd::Map(curr_joint_pos.data(), boost::numeric_cast<Eigen::Index>(curr_joint_pos.size()));
     const Eigen::VectorXd curr_waypoint_rad = curr_waypoint_deg * (M_PI / 180.0);
-    waypoints.insert(waypoints.begin(), curr_waypoint_rad);
+    if (!curr_waypoint_rad.isApprox(waypoints.front(), 0.0001)) {
```
acmorrow, line 643: - Is it a precondition for this function that `waypoints` be non-empty? It doesn't seem to be enforced.
- Constant for epsilon, should be same as one above I think.
- I think using the `else if` makes it harder to think about. If you make it a standalone conditional it reads better to me:
```
if (!curr_waypoint ... ) {
    waypoints.insert(...);
}

if (waypoints.size() == 1) {
    VIAM_SDK_LOG << ...
    return
}
```

src/viam/ur/module/ur_arm.cpp
```diff
@@ -507,7 +512,10 @@ void URArm::keep_alive() {
     }
     VIAM_SDK_LOG(info) << "keep_alive thread terminating";
 }
-
+bool is_same_position(const Eigen::VectorXd vec_a, const Eigen::VectorXd vec_b) {
```
acmorrow: debugging?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -552,10 +565,15 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
...
         const auto start = boost::numeric_cast<decltype(waypoints)::difference_type>(segments[i]);
+
         const auto end = boost::numeric_cast<decltype(waypoints)::difference_type>(segments[i + 1] + 1);
+
         const std::list<Eigen::VectorXd> positions_subset(waypoints.begin() + start, waypoints.begin() + end);
+
         const Trajectory trajectory(Path(positions_subset, 0.1), max_velocity, max_acceleration);
+
         trajectory.outputPhasePlaneTrajectory();
+
```
acmorrow: Can we collapse this back together? I have changes here that will conflict annoyingly otherwise.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -573,8 +591,16 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
         if (std::isinf(duration)) {
             throw std::runtime_error("trajectory.getDuration() was infinite");
         }
+        if (std::isnan(duration)) {
+            throw std::runtime_error("trajectory.getDuration() was nan");
+        }
+        if (duration > 600) {  // if the duration is longer than 10 minutes
```
acmorrow, line 705: This feels arbitrary. Should it be a configuration parameter?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -573,8 +591,16 @@ void URArm::move(std::vector<Eigen::VectorXd> waypoints, std::chrono::millisecon
         if (std::isinf(duration)) {
             throw std::runtime_error("trajectory.getDuration() was infinite");
         }
+        if (std::isnan(duration)) {
```
acmorrow, line 702: You could maybe collapse to `isfinite`, though you would lose the distinction between the cases unless you used `fpclassify` to synthesize the message.

acmorrow: To be clear, by collapse i mean "combine the calls to `isinf` and `isnan`"

src/viam/ur/module/ur_arm.cpp
JohnN193: ah yeah this was testing, had another round of cleanup to go through when I first put up the pr

src/viam/ur/module/ur_arm.cpp
JohnN193: def arbitrary, came up with this number when talking to rand. I can go ahead and make it configurable

acmorrow: Let's ticket it, TODO it, and leave this as is for purposes of this review.

src/viam/ur/module/ur_arm.cpp
JohnN193: yesish, `move_to_joint_positions` and `move_through_joint_positions` should actually validate that  waypoints are non empty, but I don't know if a viam-server can send the module an empty set of waypoints

acmorrow (approved):
LGTM

PR #64, opened 2025-06-26
[RSDK-11064] CI work: check lint,fmt and run tests

mattjperez:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
+name: Universal Robots CI Pipeline
```
mattjperez, line 1: main file which coordinates the others

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  test:
+    name: Run CI Tests
+    needs: [format]
+    uses: ./.github/workflows/test.yml
+
+  publish:
+    name: Publish
+    needs: [format,test]
+    if: startsWith(github.ref, 'refs/tags/v')
```
mattjperez, line 31: only run publish on tagged releases

.github/workflows/publish.yml
```diff
@@ -21,7 +22,7 @@ jobs:
         steps:
             - uses: actions/checkout@v4
               with:
-                submodules: 'true'
+                submodules: true
```
mattjperez, line 25: verified in other workflows this doesn't need to be a string

Makefile
```diff
@@ -1,27 +1,30 @@
-.PHONY: module.tar.gz ur5e-sim format test run-clang-tidy run-clang-check clean clean-all docker docker-build docker-amd64 docker-upload appimages docker-arm64-ci docker-amd64-ci
-default: build/universal-robots
+.PHONY: module.tar.gz ur5e-sim format format-check test run-clang-tidy run-clang-check clean clean-all docker docker-build docker-amd64 docker-upload appimages docker-arm64-ci docker-amd64-ci
+default: build
 
 # format the source code
 format:
 	ls src/viam/ur/module/*.*pp | xargs clang-format-19 -i --style=file
 
-build:
```
mattjperez, line 8: the cmake config generation will instantiate the `build` folder already.
confirmed this test passing in current PR.

Makefile
```diff
@@ -1,27 +1,30 @@
...
-	mkdir build
+format-check:
+	ls src/viam/ur/module/*.*pp | xargs clang-format-19 -i --style=file --dry-run --Werror
```
mattjperez, line 9: not sure if I need `--dry-run` but figured it wouldn't hurt

Makefile
```diff
@@ -1,27 +1,30 @@
...
+	cmake --build build --target all -- -j4
+	cmake --install build --prefix build/install/usr
+
+test: build
 	./build/universal-robots-test
 
 run-clang-tidy:
-	clang-tidy-19 --config-file ./.clang-tidy -p build ./src/viam/ur/module/*.cpp --header-filter=".*/viam/ur/module/.*"
+	clang-tidy-19 --config-file ./.clang-tidy -p build \
+	./src/viam/ur/module/*.cpp --header-filter=".*/viam/ur/module/.*"
```
mattjperez, line 23: just a driveby readability change

Makefile
```diff
@@ -72,13 +75,13 @@ endef
 # Targets for building AppImages
 appimage-arm64: export OUTPUT_NAME = universal-robots
 appimage-arm64: export ARCH = aarch64
-appimage-arm64: format build/universal-robots
+appimage-arm64: format build
```
mattjperez, line 78: we're not building anything other than universal-robots, so seemed redundant

.github/workflows/publish.yml
```diff
@@ -1,11 +1,12 @@
 # see https://github.com/viamrobotics/upload-module for help
+name: Publish Module
+
 on:
-    push:
-    release:
-      types: [published]
```
mattjperez, line 5: I'm not sure if this is needed

JohnN193:
> > confirmed the publish job is skipped, but would prefer not to see it at all
> 
> we could make a publish workflow that only runs on releases instead of tying it to `ci.yaml`. thats the standard setup we use for most other modules https://github.com/viam-modules/frame-calibration/blob/main/.github/workflows/deploy.yml
> 
> we can't use build-action for this module so would still need all the build steps, which may not be worth

example with build action https://github.com/viam-modules/raspberry-pi/blob/main/.github/workflows/deploy.yml

JohnN193 (approved):
lgtm, just a nit/q

mattjperez:
> > > confirmed the publish job is skipped, but would prefer not to see it at all
> > 
> > 
> > we could make a publish workflow that only runs on releases instead of tying it to `ci.yaml`. thats the standard setup we use for most other modules https://github.com/viam-modules/frame-calibration/blob/main/.github/workflows/deploy.yml
> > we can't use build-action for this module so would still need all the build steps, which may not be worth
> 
> example with upload https://github.com/viam-modules/raspberry-pi/blob/main/.github/workflows/deploy.yml

there's the possibility that these changes will need quick follow-ups to get the behavior right for releasing (since we can never know the exact behavior on main until it's actually merged). 

I'm fine with either way now; 1) following the current publish convention directly or 2) trying this out first and switching back if needed and adjusting soon after.

.github/workflows/publish.yml
mattjperez: @acmorrow actually, we might want this. might simplify your release workflow a bit.

It wouldn't be checking for version tags though using the regex that checks if it starts with `vX.Y.Z`

.github/workflows/format.yml
```diff
@@ -0,0 +1,25 @@
+name: Format and Linter Checks
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  checks:
```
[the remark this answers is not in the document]
mattjperez: That's just an arbitrary job name, not GH action keyword

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  test:
+    name: Run CI Tests
+    needs: [format]
+    uses: ./.github/workflows/test.yml
+
+  publish:
+    name: Publish
+    needs: [format,test]
+    if: startsWith(github.ref, 'refs/tags/v')
+    uses: ./.github/workflows/publish.yml
```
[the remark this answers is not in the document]
mattjperez: 5 checkouts aren't really avoidable since we need
- 1x format/checks (x86)
- 2x tests (x86 and arm)
- 2x publish (x86 and arm)

I've moved everything to the same file. gonna review it with abe later today to see are any other obvious improvements or ways to get closer to what you're saying. 

The current setup seems the most straightforward, and checkout ideally is a quick step

PR #66, opened 2025-06-27
[RSDK-11087] Remove needless waypoint copying and intermediates and fixup segmentation

acmorrow:


src/viam/ur/module/ur_arm.cpp
```diff
@@ -51,14 +52,24 @@ std::chrono::milliseconds unix_now_ms() {
     return chrono::duration_cast<chrono::milliseconds>(chrono::system_clock::now().time_since_epoch());
 }
 
+template <typename T>
+[[nodiscard]] constexpr decltype(auto) degrees_to_radians(T&& degrees) {
```
acmorrow, line 56: These are more complicated than you might expect because I want `T` to be able to be an Eigen expression template type, and to return an expression template representing the deferred conversion. Need to double check that `decltype(auto)` is right here, but I think it is.

acmorrow (commented):
There are a number of other simplifications I want to make, but this state was a good stopping point.

JohnN193 (approved):

src/viam/ur/module/ur_arm.cpp
```diff
@@ -496,23 +507,22 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
     }
     // TODO: use options
     if (!positions.empty()) {
-        std::vector<Eigen::VectorXd> waypoints;
-        for (auto position : positions) {
-            const Eigen::VectorXd next_waypoint_deg =
-                Eigen::VectorXd::Map(position.data(), boost::numeric_cast<Eigen::Index>(position.size()));
-            const Eigen::VectorXd next_waypoint_rad = next_waypoint_deg * (M_PI / 180.0);  // convert from radians to degrees
+        std::list<Eigen::VectorXd> waypoints;
```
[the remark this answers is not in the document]
acmorrow: This is just a default constructed list, which is empty. It'd mean the same thing, but more typing.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -640,36 +650,50 @@ void URArm::move_(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseco
...
+    // no cusps were identified the only) segment. If one or more cusp
+    // points were identified, the waypoints list will always have at
+    // least two residual waypoints, since the last waypoint is never
+    // examined, and the splice call never removes the waypoint being
+    // visited.
+    //
+    // NOTE: This assumes waypoints have been de-duplicated to avoid
+    // zero-length segments that would cause numerical issues in
+    // normalized() calculations.
+    std::vector<decltype(waypoints)> segments;
```
mattjperez, line 680: Same here with initialization

src/viam/ur/module/ur_arm.cpp
```diff
@@ -640,36 +650,50 @@ void URArm::move_(std::vector<Eigen::VectorXd> waypoints, std::chrono::milliseco
...
+    for (auto where = next(begin(waypoints)); where != prev(end(waypoints)); ++where) {
+        const auto segment_ab = *where - *prev(where);
+        const auto segment_bc = *next(where) - *where;
+        const auto dot = segment_ab.normalized().dot(segment_bc.normalized());
         if (std::fabs(dot + 1.0) < 1e-3) {
-            segments.push_back(i - 1);
+            segments.emplace_back();
+            segments.back().splice(segments.back().begin(), waypoints, waypoints.begin(), where);
```
mattjperez, line 687: Not sure if I'm reading this part right. Why are we starting `waypoints.begin()` here?

acmorrow: We've been walking along. We found a point we want to segment on. So, everything we've walked before (e.g. `begin()` up to but not including `where`) should get spliced into the new list. We then copy in the cusp point to the new list, since we want the last point of that segment to be the cusp point.

So, after we have done all that, where does `waypoints.begin()` point? It points to the cusp point. And that's correct. Because we want the next segment to start from that point too. Visually:

```
waypoints: a -> b -> c -> d -> e -> end
                   b              w             
```

Here we have a four element waypoints list. `b` for begin points to `a`. We are looking at `c` and determine it is a cusp point. We do our splice and copy dance, and now things look like:

```
first segment: a -> b -> c -> end

waypoints: c -> d -> e -> end
                   b
                   w
```

And iteration continues. Does that help clear it up?

src/viam/ur/module/ur_arm.cpp
acmorrow: Default initialization for containers is fine.

PR #69, opened 2025-06-27
[RSDK-11092] bump ur client lib 2.0.0->2.1.0

mattjperez:
need to do tests on real arm before merging


acmorrow (approved):
LGTM, but yes, please test on an actual arm before merging. We are due for that recently anyway with some other recent changes on main that probably only have simulator validation so far. So this is a good checkpoint for physical.

PR #73, opened 2025-07-01
[RSDK-11090] use uniform sampling

JohnN193:
switches the way we sample trajectory generation to ensure we always include the full duration. 

previously when we always sampled at 5Hz there would be a portion of the trajectory that had a smaller step size. this would cause issues if the step size was too small (~0.001 seconds) by making the arm think we sent a 0 duration trajectory. 

tested on a ur5 with & without motion and generating a segment. still need to write a unit test for the sampler.


src/viam/ur/module/ur_arm.cpp
```diff
@@ -631,6 +623,38 @@ void URArm::keep_alive_() {
...
+    // Calculate the actual step size
+    const double step = duration / static_cast<double>((num_samples - 1));
+
+    // Generate samples by evaluating f at each time point
+    for (std::size_t i = 0; i < num_samples - 1; ++i) {
+        samples.push_back(f(static_cast<double>(i) * step, step));
+    }
+
+    // Ensure the last sample uses exactly the duration
+    samples.push_back(f(duration, step));
```
JohnN193, line 653: so looking at this again, I actually am not sure we even need this final step anymore. the previous samples should have a total duration that reaches the end of the trajectory, so adding this here is the equivalent to saying "do one more step with zero velocity"

its probably fine but if other people have thoughts lmk

src/viam/ur/module/ur_arm.cpp
```diff
@@ -133,28 +133,51 @@ auto make_scope_guard(Callable&& cleanup) {
...
+    if (num_samples == 1) {
+        samples.push_back(f(0.0, duration));
         return;
     }
+
+    // std::vector<trajectory_sample_point> samples;
+    samples.reserve(num_samples);
+
+    // Calculate the actual step size
+    const double step = duration / static_cast<double>((num_samples - 1));
```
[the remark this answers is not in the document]
JohnN193: num_samples will always be at least 2. updating the code to make that more apparent

acmorrow (changes requested):
I like where this is going. I got a little picky about things because this seemed like a good opportunity.

src/viam/ur/module/test.cpp
```diff
@@ -46,20 +46,10 @@ BOOST_AUTO_TEST_CASE(test_write_waypoints_to_csv) {
...
-    };
-    const std::vector<vector6d_t> p_v = {
-        {1.2, 2, 3, 4, 5, 6},
-        {4.2, 2, 3, 4, 5, 6},
-        {7.1, 2, 3, 4, 5, 6},
-        {10.1, 2, 3, 4, 5, 6},
-    };
-    // the time vector is provided in relative time
-    const std::vector<float> time = {1.2F, 0.8F, 1, 1};
+    const std::vector<trajectory_sample_point> samples = {{vector6d_t{1.1, 2, 3, 4, 5, 6}, vector6d_t{1.2, 2, 3, 4, 5, 6}, 1.2F},
```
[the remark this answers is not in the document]
JohnN193: when I was first setting up tests this was yelling at me, but that was prob because of a different bug

JohnN193:
@acmorrow @mattjperez added tests, they really suck atm. lmk what else I can do with them.

src/viam/ur/module/ur_arm.hpp
```diff
@@ -13,10 +13,13 @@
...
 using namespace urcl;
 
-void write_trajectory_to_file(const std::string& filepath,
-                              const std::vector<vector6d_t>& p_p,
-                              const std::vector<vector6d_t>& p_v,
-                              const std::vector<float>& time);
+struct trajectory_sample_point {
+    vector6d_t p;
+    vector6d_t v;
+    float timestep;
```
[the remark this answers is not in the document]
acmorrow: Oh, I see. Yeah, that's annoying. Keep it as it is then.

src/viam/ur/module/test.cpp
```diff
@@ -13,8 +13,56 @@ Eigen::VectorXd makeVector(std::vector<double> data) {
...
+        sampling_func(test_samples, test_duration_sec, test_freq_hz, [](const double t, const double step) {
+            return trajectory_sample_point{{t, 0, 0, 0, 0, 0}, {t * step, 0, 0, 0, 0, 0}, boost::numeric_cast<float>(step)};
+        });
+        BOOST_CHECK_EQUAL(test_samples.size(), 2);
+        BOOST_CHECK_EQUAL(test_samples[0].p[0], 0);
+        BOOST_CHECK_EQUAL(test_samples[0].v[0], 0);
+        BOOST_CHECK_EQUAL(test_samples[0].timestep, boost::numeric_cast<float>(test_duration_sec));
+        BOOST_CHECK_EQUAL(test_samples[1].p[0], test_duration_sec);
+        BOOST_CHECK_EQUAL(test_samples[1].v[0], test_duration_sec * test_duration_sec);
+        BOOST_CHECK_EQUAL(test_samples[1].timestep, boost::numeric_cast<float>(test_duration_sec));
```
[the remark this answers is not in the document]
acmorrow: Thought about it a bit and I see what you are saying. I want @randhid's opinion on this, but, I think it is actually the t(0) sample that we don't want. The first waypoint of each segment is the current joint position, by construction. If there is only one segment, then the first waypoint is the position of the arm at the start of the move call, since we inserted it as such. If we segmented, then the prior argument holds for the first segment, and the first waypoint of each subsequent segment is equal to the last waypoint of the prior segment, so, again, we are already "there". If this logic projects through trajectory planning, I'd expect moving from the current arm position to t(0) of each segment's computed trajectory to be a always be a no-op, and we shouldn't be allocating `step` units of time for it to happen. It should be pretty easy to validate this by looking at the trajectory logs. If that's the case, we shouldn't be sampling at t(0) at all. And then, without that sample, you don't end up with the extra time in the degenerate case as you do now. So that will imply some small changes to the sampling function.

src/viam/ur/module/test.cpp
```diff
@@ -13,8 +13,57 @@ Eigen::VectorXd makeVector(std::vector<double> data) {
     return Eigen::Map<Eigen::VectorXd>(data.data(), boost::numeric_cast<Eigen::Index>(data.size()));
 }
 
+// trajectory_sample_point
```
acmorrow, line 16: remove

src/viam/ur/module/test.cpp
```diff
@@ -13,8 +13,57 @@ Eigen::VectorXd makeVector(std::vector<double> data) {
...
+    {
+        std::vector<trajectory_sample_point> test_samples = {};
+        const double test_duration_sec = 0;
+        const double test_freq_hz = 0.5;  // 1 sample every 2 seconds
+
+        BOOST_CHECK_THROW(sampling_func(test_samples,
+                                        test_duration_sec,
+                                        test_freq_hz,
+                                        [](const double t, const double step) {
+                                            return trajectory_sample_point{vector6d_t{t, 0, 0, 0, 0, 0},
```
acmorrow, line 59: You could just fail the test in the lambda, since it should never be called.

acmorrow (changes requested):

src/viam/ur/module/ur_arm.hpp
```diff
@@ -13,10 +13,39 @@
...
+    }
+
+    // Calculate the number of samples needed. this will always be at least 2.
+    const auto num_samples = static_cast<std::size_t>(std::ceil(putative_samples) + 1);
+
+    // Calculate the actual step size
+    const double step = duration_sec / static_cast<double>((num_samples - 1));
+
+    // Generate samples by evaluating f at each time point
+    for (std::size_t i = 0; i < num_samples - 1; ++i) {
```
acmorrow, line 40: If we end up deciding to take my suggestion to skip the t(0) sample, just change this to `std::size_t i = 1` and I think you get exactly what you want.

src/viam/ur/module/test.cpp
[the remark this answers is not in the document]
JohnN193: Gonna do an experiment with t(0) and t(duration) to see which one actually moves the arm, or if they both do

src/viam/ur/module/test.cpp
```diff
@@ -13,8 +13,57 @@ Eigen::VectorXd makeVector(std::vector<double> data) {
...
+    {
+        std::vector<trajectory_sample_point> test_samples = {};
+        const double test_duration_sec = 5;
+        const double test_freq_hz = 5;
+        // we are not testing the behavior of the trajectory generation, so we create a simple lambda function.
+        // this allows us to get a good idea of what sampling_func is doing
+        sampling_func(test_samples, test_duration_sec, test_freq_hz, [](const double t, const double step) {
+            return trajectory_sample_point{{t, 0, 0, 0, 0, 0}, {t * step, 0, 0, 0, 0, 0}, boost::numeric_cast<float>(step)};
+        });
+        BOOST_CHECK_EQUAL(test_samples.size(), 26);
```
mattjperez, line 30: could probably make this `BOOST_CHECK_EQUAL(test_samples.size(), test_duration_sec*test_freq_hz + 1)` so it doesn't seem like a magic number. will be easy to adjust too pending on if we include t0

src/viam/ur/module/test.cpp
```diff
@@ -13,8 +13,57 @@ Eigen::VectorXd makeVector(std::vector<double> data) {
...
+    }
+    // check for durations smaller than the sampling frequency
+    {
+        std::vector<trajectory_sample_point> test_samples = {};
+        const double test_duration_sec = 1;
+        const double test_freq_hz = 0.5;  // 1 sample every 2 seconds
```
mattjperez, line 36: not sure if this is necessarily a good combo for checking 'durations smaller than sampling frequency' because it hits the lower bounds for sampling size. might be more accurate to say this 'checks for minimum sampling bounds on valid input'.

mattjperez (approved):
LGTM, thanks for walking me stuff!

acmorrow (approved):
LGTM. Ship it!

PR #76, opened 2025-07-02
[RSDK-11180] resend robot program

JohnN193:
was able to reproduce the error rand and ray saw on the ur20 and ur3e. When the arm is moved during local mode, it clears the control program that we sent via the ur driver. resending the control script when we detect the failure allows the arm to be moved again. 

additionally, this pr adds support for getting data from the arm during local mode. I saw the motion team recording joint positions by hand and i was not a fan of that. it makes the state of the thread even harder to follow though, so sorry about that.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -954,9 +954,16 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_(bool log) {
         VIAM_SDK_LOG(error) << "failed to talk to the arm, is the tablet in local mode? : " << std::string(ex.what());
     }
 
+    // check if the arm status is normal or not empty. the status will be empty if the arm is disconnected or when the arm is first switched
+    // into local mode
     if ((status.find(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL)) == std::string::npos) && !status.empty()) {
-        // the arm is currently estopped. save this state.
-        current_state_->estop.store(true);
+        VIAM_SDK_LOG(info) << "read_joint_keep_alive dashboard->commandSafetyStatus() arm status : " << status;
+        // the arm is currently estopped. if the user is not in local mode, save this state.
+        // if the user is in local mode, then we assume that they opted into this state.
+        // the UR20 will always return a stop while in manual mode, unless the three position enabling button is held down
+        if (!current_state_->local_disconnect.load()) {
+            current_state_->estop.store(true);
+        }
 
         // clear any currently running trajectory.
         current_state_->trajectory_status.store(TrajectoryStatus::k_stopped);


src/viam/ur/module/ur_arm.cpp
```diff
@@ -897,34 +890,38 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_(bool log) {
...
+                        throw std::runtime_error(
+                            "read_joint_keep_alive driver->sendRobotProgram() returned false when attempting to reconnect to the arm");
+                    }
+                }
 
-            if (!current_state_->dashboard->commandIsInRemoteControl()) {
-                return UrDriverStatus::DASHBOARD_FAILURE;
-            }
+                // turn communication with the RTDE client back on so we can receive data from the arm
+                current_state_->driver->startRTDECommunication();
```
JohnN193, line 917: don't like how this part of the code keeps growing with random reset commands. I don't think I can simplify this any further though

src/viam/ur/module/ur_arm.cpp
```diff
@@ -948,13 +945,25 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_(bool log) {
...
+    // check if the arm status is normal or not empty. the status will be empty if the arm is disconnected or when the arm is first switched
+    // into local mode
+    if ((status.find(urcl::safetyStatusString(urcl::SafetyStatus::NORMAL)) == std::string::npos) && !status.empty()) {
+        VIAM_SDK_LOG(info) << "read_joint_keep_alive dashboard->commandSafetyStatus() arm status : " << status;
+        // the arm is currently estopped. if the user is not in local mode, save this state.
+        // if the user is in local mode, then we assume that they opted into this state.
+        // the UR20 will always return a stop while in manual mode, unless the three position enabling button is held down
+        if (!current_state_->local_disconnect.load()) {
+            current_state_->estop.store(true);
+        }
```
JohnN193, line 966: also not a huge fan of this. the other "easy" option is to always run through the estop recovery if an estop is detected in local mode. 

the issue I had with that was the ur20 basically always reports an estop when switching into local mode, and that could get pretty tedious for the motion team while testing

src/viam/ur/module/ur_arm.cpp
```diff
@@ -897,34 +890,38 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_(bool log) {
...
+                current_state_->driver->resetRTDEClient(current_state_->appdir + k_output_recipe, current_state_->appdir + k_input_recipe);
+                // reset the primary client so the driver is aware the arm is back in remote mode
+                // this has to happen, otherwise when an estop is triggered the arm will return error code C210A0
+                current_state_->driver->stopPrimaryClientCommunication();
+                current_state_->driver->startPrimaryClientCommunication();
+
+                // check if we can still send trajectories. if not we need to send the robot program again.
+                if (!current_state_->driver->isReverseInterfaceConnected() || !current_state_->driver->isTrajectoryInterfaceConnected()) {
+                    // send control script again to complete the restart
+                    if (!current_state_->driver->sendRobotProgram()) {
```
JohnN193, line 910: Moving the arm while in local mode removes the control script, so you have to resend the robot program.

This is the line that fixed the issue seen on the ur20. I actually had it while testing on the ur5e, but removed it when it appeared to do nothing. 

The majority of the other changes in this code are for enabling data while in local mode

src/viam/ur/module/ur_arm.cpp
```diff
@@ -897,34 +890,38 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_(bool log) {
     std::string status;
     try {
         if (current_state_->local_disconnect.load()) {
-            // sleep just so we don't spam
-            // this may be unnecessary
-            std::this_thread::sleep_for(k_estop_delay);
+            if (current_state_->dashboard->commandIsInRemoteControl()) {
```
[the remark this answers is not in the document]
acmorrow: It's a good suggestion, but I'd like to avoid this fix becoming a reorg of the code. Ideally, we get all the necessary functionality and features and fixes in, even if we don't love the organization. Then we can look at the whole thing and figure out how to structure it better. I'm actually working on some of that right now, though I'm hoping to mostly leave this logic alone initially.

acmorrow (approved):
This LGTM. I agree with your concerns about the worker thread but agree that a better design shouldn't be part of this review.

PR #80, opened 2025-07-08
[RSDK-10757] handle recovery if the arm was off or disconnected

JohnN193:
this pr handles the case where the arm has been turned off, unplugged, or lost internet connection. This should work with both sim and hardware, though I tested significantly less with sim.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -207,6 +207,9 @@ struct URArm::state_ {
     std::atomic<bool> estop{false};
     std::atomic<bool> local_disconnect{false};
 
+    // variables derived from arm state on configuration
+    bool is_sim;
+
     std::mutex output_csv_dir_path_mu;
     // specified through VIAM_MODULE_DATA environment variable
     std::string output_csv_dir_path;
@@ -319,7 +322,7 @@ void URArm::startup_(const Dependencies&, const ResourceConfig& cfg) {
 
     // connect to the robot dashboard
     current_state_->dashboard.reset(new DashboardClient(current_state_->host));
-    if (!current_state_->dashboard->connect()) {
+    if (!current_state_->dashboard->connect(1)) {
         throw std::runtime_error("couldn't connect to dashboard");
     }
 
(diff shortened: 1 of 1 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -357,16 +360,22 @@ void URArm::startup_(const Dependencies&, const ResourceConfig& cfg) {
...
-                                                true,  // headless mode
-                                                nullptr};
+    urcl::UrDriverConfiguration ur_cfg = {current_state_->host,
+                                          current_state_->appdir + k_script_file,
+                                          current_state_->appdir + k_output_recipe,
+                                          current_state_->appdir + k_input_recipe,
+                                          &reportRobotProgramState,
+                                          true,  // headless mode
+                                          nullptr};
+    ur_cfg.socket_reconnect_attempts = 1;
```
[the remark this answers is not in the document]
mattjperez: if we're gonna do this, we might as well do all the fields like this. can pull it into a private method to build the config.
https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/289c747850e2bd0c6262a39310b2ae56593baee6/include/ur_client_library/ur/ur_driver.h#L53

```cpp
auto ur_cfg = urcl::UrDriverConfiguration{};
ur_cfg.robot_ip = std::move(current_state_->host);
ur_cfg.script_file = ...;
ur_cfg.output_recipe_file = ...;
ur_cfg.input_recipe_file = ...;
ur_cfg.handle_program_state = &reportRobotProgramState;
ur_cfg.headless_mode = true;
ur_cfg.socket_reconnect_attempts = 1;
```

src/viam/ur/module/ur_arm.cpp
```diff
@@ -894,6 +903,18 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_(bool log) {
...
+                std::string robot_mode;
+                if (!current_state_->dashboard->commandRobotMode(robot_mode)) {
+                    throw std::runtime_error("read_joint_keep_alive dashboard->commandRobotMode() failed to retrieve robot mode");
+                }
+
+                // if the robot is not running, we want to restart the arm fully to make sure we are ready
+                // we will cheat by going into the estop code
+                if (robot_mode.find(urcl::robotModeString(urcl::RobotMode::RUNNING)) == std::string::npos) {
+                    // we are not ready to run, go into the estop code to try to be ready
+                    current_state_->estop.store(true);
```
mattjperez, line 915: how does setting a bool make the robot try to be ready? is this depending on `read_joint_keepalive` thread to restart underlying connections?

JohnN193: yeah, in the [handle estop chunk of code](https://github.com/viam-modules/universal-robots/blob/980e125ae93887d746e3dc51981b99967c3526f0/src/viam/ur/module/ur_arm.cpp#L971-L1000) we go through the process of ensuring the arm is turned on and sending the control script that we want to use. 

rather than redo all of that here I cheat by setting the flag

JohnN193: if status is empty or not normal then the arm is currently estopped, so we don't want to turn the arm back on

mattjperez:
@JohnN193 I confirmed that the behavior works on simulator, able to use a script again when a robot is turned off, then on, then script is ran.

However, I saw that using a test script when the simlator arm is off hangs. Is this behavior expected?

Otherwise LGTM.

src/viam/ur/module/ur_arm.cpp
acmorrow: https://viam.atlassian.net/browse/RSDK-11279

JohnN193:
> However, I saw that using a test script when the simlator arm is off hangs/blocks. Is this behavior expected?
> 

this is not expected and not the behavior that I see

mattjperez:
> > However, I saw that using a test script when the simlator arm is off hangs/blocks. Is this behavior expected?
> 
> this is not expected and not the behavior that I see

discussed offline. The way I did it (by pressing off in the sim) and recovering is not covered in the current internal state machine and not typical user behavior on the actual arm. Outside the scope of this PR's goals.

acmorrow (approved):
LGTM. Still please do file an upstream issue with UR about the sim behavior. Maybe they just don't know.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -357,16 +360,22 @@ void URArm::startup_(const Dependencies&, const ResourceConfig& cfg) {
         throw std::runtime_error("couldn't release the arm brakes");
     }
 
+    // sim always reports that it is in local mode, but during local mode we cannot turn the arm off/on like we just did.
```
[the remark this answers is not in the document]
JohnN193: https://github.com/UniversalRobots/Universal_Robots_Client_Library/issues/360

PR #83, opened 2025-07-11
[RSDK-10956] allow multiple UR arms on the same robot part

gvaradarajan:
Each instantiated UR driver for a single module needs to be passed different client-side ports. This commit increments atomic integers for each port to fix this.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -457,6 +457,15 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
     ur_cfg.headless_mode = true;
     ur_cfg.socket_reconnect_attempts = 1;
 
+    ur_cfg.reverse_port = reverse_port.fetch_add(4);
+    ur_cfg.script_sender_port = script_sender_port.fetch_add(4);
+    ur_cfg.trajectory_port = script_sender_port.fetch_add(4);
+    ur_cfg.script_command_port = script_command_port.fetch_add(4);
+    VIAM_SDK_LOG(debug) << "using reverse_port " << ur_cfg.reverse_port;
+    VIAM_SDK_LOG(debug) << "using script_sender_port " << ur_cfg.script_sender_port;
+    VIAM_SDK_LOG(debug) << "using trajectory_port " << ur_cfg.trajectory_port;
+    VIAM_SDK_LOG(debug) << "using script_command_port " << ur_cfg.script_command_port;
+
     current_state_->driver.reset(new UrDriver(ur_cfg));
 
     // define callback function to be called by UR client library when trajectory state changes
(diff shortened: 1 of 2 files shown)


mattjperez (commented):
Seems straightforward. Just a couple questions.

npmenard (changes requested):
I think you should use one atomic to protect ports so that there is no risk of crossing port range if arm are created by different threads

src/viam/ur/module/ur_arm.hpp
```diff
@@ -54,6 +54,22 @@ std::string trajectory_filename(const std::string& path, const std::string& unix
...
+struct ports {
+    uint32_t reverse_port;
+    uint32_t script_sender_port;
+    uint32_t trajectory_port;
+    uint32_t script_command_port;
+};
+
+// We need to requisition different ports for each independent URArm instance, otherwise they will all try
+// to use the same ports and only one of them will work.
+inline ports new_ports() {
```
npmenard, line 67: nit: auto

npmenard (approved):
LGTM some nits

src/viam/ur/module/ur_arm.cpp
```diff
@@ -457,6 +457,18 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
+        current_ports_ = new_ports();
+    }
+    ur_cfg.reverse_port = current_ports_->reverse_port;
+    ur_cfg.script_sender_port = current_ports_->script_sender_port;
+    ur_cfg.trajectory_port = current_ports_->trajectory_port;
+    ur_cfg.script_command_port = current_ports_->script_command_port;
+    VIAM_SDK_LOG(debug) << "using reverse_port " << ur_cfg.reverse_port;
+    VIAM_SDK_LOG(debug) << "using script_sender_port " << ur_cfg.script_sender_port;
+    VIAM_SDK_LOG(debug) << "using trajectory_port " << ur_cfg.trajectory_port;
+    VIAM_SDK_LOG(debug) << "using script_command_port " << ur_cfg.script_command_port;
```
JohnN193, line 470: if these do not show up in the logs when debug is configured, go ahead and change them to info level

JohnN193 (changes requested):
> Yeah it'll fail in the manner that it did before, do you think it's worth making the initial port number configurable? I'm hoping this risk is minimized by picking the same very high port range (50000s) as the UR client library does
> I'm not worried because it's fixable by restarting the viam server

other people/motion can overrule me, but I do not believe restarting the viam-server is an acceptable solution. especially when it won't be obvious when a user has to restart their viam-server due to this issue

src/viam/ur/module/ur_arm.hpp
```diff
@@ -54,6 +54,22 @@ std::string trajectory_filename(const std::string& path, const std::string& unix
...
+    static std::atomic<uint32_t> port_counter(50001);
+    const uint32_t reverse_port = port_counter.fetch_add(4);
```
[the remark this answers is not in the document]
gvaradarajan: I'm curious, what is preferred about auto in these cases?

JohnN193 (approved):

mattjperez (approved):

PR #89, opened 2025-07-23
[RSDK-11307] fix unrecoverable case on reconfigure 

mattjperez:
On L1176, there is an ordered check that first checks `current_state->local disconnect` before setting the `current_state->estop` to `true`, however, in practice, it's possible for an operator to switch to manual mode *before* releasing the estop. 

This leads to a path on L1190 where we try to recovery from an estop despite being in manual mode.

Additionally, some scenarios lead to `shutdown_` being called, clearing the `current_state`, `URArm::reconfigure`'s call to `check_configured_` will throw an error because `current_state` doesn't exist.

Removing the check makes sense since we should still be able to reconfigure our module even when a device is estopped or in manual mode. There's already a check in `shutdown_` that checks for state anyway.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -544,7 +544,6 @@ void URArm::trajectory_done_cb_(const control::TrajectoryResult state) {
 
 void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
     const std::unique_lock wlock{config_mutex_};
-    check_configured_(wlock);
     VIAM_SDK_LOG(warn) << "Reconfigure called: shutting down current state";
     shutdown_(wlock);
     VIAM_SDK_LOG(warn) << "Reconfigure called: configuring new state";
@@ -1188,7 +1187,7 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_inner_(bool log) {
 
     } else {
         // the arm is in a normal state.
-        if (current_state_->estop.load()) {
+        if (current_state_->estop.load() && !current_state_->local_disconnect.load()) {
             // if the arm was previously estopped, attempt to recover from the estop.
             // We should not enter this code without the user interacting with the arm in some way(i.e. resetting the estop)
             try {


src/viam/ur/module/ur_arm.cpp
```diff
@@ -1188,7 +1187,7 @@ URArm::UrDriverStatus URArm::read_joint_keep_alive_inner_(bool log) {
 
     } else {
         // the arm is in a normal state.
-        if (current_state_->estop.load()) {
+        if (current_state_->estop.load() && !current_state_->local_disconnect.load()) {
```
JohnN193, line 1190: whoops

src/viam/ur/module/ur_arm.cpp
```diff
@@ -544,7 +544,6 @@ void URArm::trajectory_done_cb_(const control::TrajectoryResult state) {
 
 void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
     const std::unique_lock wlock{config_mutex_};
-    check_configured_(wlock);
```
JohnN193, line 547: if `configure_` throws then we have no `current_state_`, which causes the resource to fail without being able to recover

JohnN193 (approved):

src/viam/ur/module/ur_arm.cpp
acmorrow: This was by design. My understanding was that if a resource failed reconfigure, that it would be removed from the resource graph and destroyed. So the only way that something could be `reconfigure`d was if it had either 1) been created, or 2) had already been `reconfigure'd OK. Either way, there should always be a state that is being reconfigured.

So, I'm not sure about this one. But I'm OK with being wrong, as long as there is a good explanation for how the resource can end up in a "failed to configure state" such that it would see another `reconfigure` call where `check_configured_` would fail.

src/viam/ur/module/ur_arm.cpp
acmorrow: Basically, don't attempt to recover from estop if we are already in local mode?

mattjperez: yup. We don't really want to be interacting with the machine while in local mode. So should probobably switch the order of these to highlight that context and short-circuit when possible

acmorrow:
@mattjperez and @JohnN193 - Any updates to this review?

JohnN193:
> @mattjperez and @JohnN193 - Any updates to this review?

Sorry, meant to go dig into rdk to see when we call reconfigure.

src/viam/ur/module/ur_arm.cpp
acmorrow: I checked with @randhid, and this was probably my misunderstanding. I had believed that if you failed a `reconfigure` that the node would be removed from the resource graph. This appears not to be true: it will continue to attempt to reconfigure the resource. So, my check was too restrictive, and I agree with removing it. Especially if you have the empirical answer that it really does happen.

PR #93, opened 2025-08-06
Add threshold to reject move request

npmenard:
This PR adds a configuration options `reject_move_request_threshold_deg` to reject motion request in the event the first position of the waypoint of the request differs by more than the set threshold from the current position.

I elected to use the Lp infinity norm (Chebyshev distance) that gives us the maximum absolute value of the joint difference. In my opinion this better than using the classic euclidean distance since in this case small differences in each joint might yield a value above the threshold even if the two joint positions are close to each other.

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -16,6 +16,7 @@
 #include <boost/io/ostream_joiner.hpp>
 #include <boost/numeric/conversion/cast.hpp>
 #include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/algorithm.hpp>
 
 #include <ur_client_library/types.h>
 #include <ur_client_library/ur/dashboard_client.h>
@@ -588,18 +589,12 @@ void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
 std::vector<double> URArm::get_joint_positions(const ProtoStruct&) {
     const std::shared_lock rlock{config_mutex_};
     check_configured_(rlock);
-    auto curr_joint_position_rad = get_joint_positions_rad(rlock);
-    std::vector<double> joint_position_degree;
-    joint_position_degree.reserve(curr_joint_position_rad.size());
-    std::transform(std::make_move_iterator(curr_joint_position_rad.begin()),
-                   std::make_move_iterator(curr_joint_position_rad.end()),
-                   std::back_inserter(joint_position_degree),
-                   radians_to_degrees<double>);
-
-    return joint_position_degree;
+    auto joint_rads = get_joint_positions_rad_(rlock);
+    auto joint_position_degree = joint_rads | boost::adaptors::transformed(radians_to_degrees<const double&>);
+    return {std::begin(joint_position_degree), std::end(joint_position_degree)};
 }
 
-vector6d_t URArm::get_joint_positions_rad(const std::shared_lock<std::shared_mutex>&) {
+vector6d_t URArm::get_joint_positions_rad_(const std::shared_lock<std::shared_mutex>&) {
     const std::lock_guard guard{current_state_->state_mutex};
     if (current_state_->last_driver_status == UrDriverStatus::READ_FAILURE) {
         // TODO(RSDK-11295): provide more context
@@ -834,7 +829,7 @@ void URArm::move_(std::shared_lock<std::shared_mutex> config_rlock, std::list<Ei
 
     // get current joint position and add that as starting pose to waypoints
     VIAM_SDK_LOG(info) << "move: get_joint_positions start " << unix_time;
-    auto curr_joint_pos = get_joint_positions_rad(our_config_rlock);
+    auto curr_joint_pos = get_joint_positions_rad_(our_config_rlock);
     VIAM_SDK_LOG(info) << "move: get_joint_positions end " << unix_time;
     auto curr_joint_pos_rad = Eigen::Map<Eigen::VectorXd>(curr_joint_pos.data(), curr_joint_pos.size());
 
@@ -845,15 +840,12 @@ void URArm::move_(std::shared_lock<std::shared_mutex> config_rlock, std::list<Ei
             std::stringstream err_string;
 
             err_string << "rejecting move request : difference between starting trajectory position [(";
-            std::transform(std::make_move_iterator(waypoints.front().cbegin()),
-                           std::make_move_iterator(waypoints.front().cend()),
-                           boost::io::make_ostream_joiner(err_string, ", "),
-                           radians_to_degrees<const double>);
+            auto first_waypoint = waypoints.front();
+            boost::copy(first_waypoint | boost::adaptors::transformed(radians_to_degrees<const double&>),
+                        boost::io::make_ostream_joiner(err_string, ", "));
             err_string << ")] and joint position [(";
-            std::transform(std::make_move_iterator(curr_joint_pos_rad.cbegin()),
-                           std::make_move_iterator(curr_joint_pos_rad.cend()),
-                           boost::io::make_ostream_joiner(err_string, ", "),
-                           radians_to_degrees<const double>);
+            boost::copy(curr_joint_pos_rad | boost::adaptors::transformed(radians_to_degrees<const double&>),
+                        boost::io::make_ostream_joiner(err_string, ", "));
             err_string << ")] is above threshold " << radians_to_degrees(delta_pos.lpNorm<Eigen::Infinity>()) << " > "
                        << radians_to_degrees(*current_state_->reject_move_request_threshold_rad);
             VIAM_SDK_LOG(error) << err_string.str();
--- a/src/viam/ur/module/ur_arm.hpp
+++ b/src/viam/ur/module/ur_arm.hpp
@@ -154,7 +154,7 @@ class URArm final : public Arm, public Reconfigurable {
 
     void keep_alive_();
 
-    vector6d_t get_joint_positions_rad(const std::shared_lock<std::shared_mutex>&);
+    vector6d_t get_joint_positions_rad_(const std::shared_lock<std::shared_mutex>&);
 
     void move_(std::shared_lock<std::shared_mutex> config_rlock, std::list<Eigen::VectorXd> waypoints, const std::string& unix_time_ms);
 


src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,6 +803,28 @@ void URArm::keep_alive_() {
 void URArm::move_(std::shared_lock<std::shared_mutex> config_rlock, std::list<Eigen::VectorXd> waypoints, const std::string& unix_time) {
     auto our_config_rlock = std::move(config_rlock);
 
+    if (reject_move_request_threshold_deg) {
+        auto joint_pos = [&]() {
```
npmenard, line 807: @acmorrow I assume capturing the lock by reference is the way to go

src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,6 +803,28 @@ void URArm::keep_alive_() {
...
+                           degrees_to_radians<double>);
+            return vector;
+        }();
+
+        auto delta_pos = (waypoints.front() - joint_pos);
+
+        if (radians_to_degrees(delta_pos.lpNorm<Eigen::Infinity>()) > *reject_move_request_threshold_deg) {
+            std::stringstream err_string;
+            err_string << "rejecting move request : difference between starting trajectory position and joint position is above threshold "
+                       << radians_to_degrees(delta_pos.lpNorm<Eigen::Infinity>()) << " > " << *reject_move_request_threshold_deg;
```
JohnN193, line 822: [opt] might be worth logging the waypoints and joint_pos too. that way users can do a lil bit of debugging without needing to go digging

src/viam/ur/module/ur_arm.cpp
```diff
@@ -372,6 +374,11 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
     current_state_->host = find_config_attribute<std::string>(cfg, "host");
     current_state_->speed.store(degrees_to_radians(find_config_attribute<double>(cfg, "speed_degs_per_sec")));
     current_state_->acceleration.store(degrees_to_radians(find_config_attribute<double>(cfg, "acceleration_degs_per_sec2")));
+    try {
+        reject_move_request_threshold_deg.emplace(find_config_attribute<double>(cfg, "reject_move_request_threshold_deg"));
```
JohnN193, line 378: validate non negative & non zero. if we want to do this in a sepatate pr since we don't validate the other config fields atm thats fine with me

JohnN193 (changes requested):
add the field to the table in the readme pls https://github.com/viam-modules/universal-robots/blob/b609bb43ae3946613770acaead595104870db983/README.md?plain=1#L31-L35

lgtm past that

src/viam/ur/module/ur_arm.cpp
JohnN193, line 378: also should we just convert this to radians here

JohnN193 (approved):

src/viam/ur/module/ur_arm.cpp
```diff
@@ -372,6 +375,18 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
     current_state_->host = find_config_attribute<std::string>(cfg, "host");
     current_state_->speed.store(degrees_to_radians(find_config_attribute<double>(cfg, "speed_degs_per_sec")));
     current_state_->acceleration.store(degrees_to_radians(find_config_attribute<double>(cfg, "acceleration_degs_per_sec2")));
+    try {
```
acmorrow, line 378: We do actually have a config validator, in `validate_config_`. Perhaps this checking belongs there. I'm not sure why it has a trailing `_`, in case you want to clean that up while you are there.

src/viam/ur/module/ur_arm.hpp
```diff
@@ -173,4 +173,5 @@ class URArm final : public Arm, public Reconfigurable {
     std::shared_mutex config_mutex_;
     std::unique_ptr<state_> current_state_;
     std::optional<ports> current_ports_;
+    std::optional<double> reject_move_request_threshold_deg;
```
acmorrow, line 176: - The `reject_move_request_threshold_deg` field should live on `struct state_` with the other configuration parameters. The `current_ports_` should be there too, but I've already cleaned that up in my tree.
- I think it might end up being simpler to store it in radians after validating it.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -372,6 +375,18 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
+        auto threshold = find_config_attribute<double>(cfg, "reject_move_request_threshold_deg");
+        if (threshold < 0 || threshold > 360) {
+            VIAM_SDK_LOG(error) << "reject_move_request_threshold_deg should be between 0 and 360, it is : " << threshold
+                                << " -  safety disabled";
+            reject_move_request_threshold_deg.reset();
+        } else {
+            reject_move_request_threshold_deg.emplace(threshold);
+        }
+    } catch (...) {
+        reject_move_request_threshold_deg.reset();
```
acmorrow, line 388: I don't think you need the `reset`s if you move this to `struct state_`, because the `struct state_` lives or dies based on what happens in `configure_`

src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,23 +811,50 @@ void URArm::keep_alive_() {
 void URArm::move_(std::shared_lock<std::shared_mutex> config_rlock, std::list<Eigen::VectorXd> waypoints, const std::string& unix_time) {
     auto our_config_rlock = std::move(config_rlock);
 
+    // get current joint position and add that as starting pose to waypoints
+    VIAM_SDK_LOG(info) << "move: get_joint_positions start " << unix_time;
+    auto curr_joint_pos_deg = get_joint_positions_(our_config_rlock);
```
acmorrow, line 816: Well, `get_joint_positions_` actually starts out in radians internally, and converts to degrees to return, but then you are converting back. The following might clear it up:

- Rename `get_joint_positions_` to `get_joint_positions_rad_`, and remove the conversion to degrees from it. You might even be able to change the return type to `vector6d` to avoid an allocation.
- Add the conversion from radians to degrees to `get_joint_positions` (the one without a trailing `_`).

Then here you can just call `get_joint_positions_rad_` and you won't need to convert back, and you can just make an Eigen::VectorXd that `Maps` over it too, I think.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,23 +811,50 @@ void URArm::keep_alive_() {
...
+    }
+
     // Capture the current movement epoch, so we can later detect if another caller
     // slipped in while we were planning.
     auto current_move_epoch = current_state_->move_epoch.load(std::memory_order_acquire);
 
     VIAM_SDK_LOG(info) << "move: start unix_time_ms " << unix_time << " waypoints size " << waypoints.size();
     const auto log_move_end = make_scope_guard([&] { VIAM_SDK_LOG(info) << "move: end unix_time " << unix_time; });
 
-    // get current joint position and add that as starting pose to waypoints
```
acmorrow, line 806: Was there a reason you effectively moved this code up above the `current_move_epoch` acquisition and the `log_move_end` guard to add the check? I think it would be more natural to insert it down here (it'd make the diff a little clearer too).

src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,23 +811,50 @@ void URArm::keep_alive_() {
...
+    if (reject_move_request_threshold_deg) {
+        auto delta_pos = (waypoints.front() - curr_joint_pos_rad);
+
+        if (radians_to_degrees(delta_pos.lpNorm<Eigen::Infinity>()) > *reject_move_request_threshold_deg) {
+            std::stringstream err_string;
+
+            err_string << "rejecting move request : difference between starting trajectory position [(";
+            std::transform(std::make_move_iterator(waypoints.front().cbegin()),
+                           std::make_move_iterator(waypoints.front().cend()),
+                           std::experimental::make_ostream_joiner(err_string, ", "),
```
acmorrow, line 833: Use `boost/io/ostream_joiner.hpp` rather than `std::experimental` which may not be reliably implemented.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,23 +811,50 @@ void URArm::keep_alive_() {
...
+    auto curr_joint_pos_rad = Eigen::VectorXd(curr_joint_pos_deg.size());
+    std::transform(std::make_move_iterator(curr_joint_pos_deg.begin()),
+                   std::make_move_iterator(curr_joint_pos_deg.end()),
+                   curr_joint_pos_rad.begin(),
+                   degrees_to_radians<double>);
+
+    if (reject_move_request_threshold_deg) {
+        auto delta_pos = (waypoints.front() - curr_joint_pos_rad);
+
+        if (radians_to_degrees(delta_pos.lpNorm<Eigen::Infinity>()) > *reject_move_request_threshold_deg) {
```
acmorrow, line 827: If you take my suggestion to store the threshold in radians you could eliminate this conversion on the check, and then only do it when reporting a violation so that the UX is consistently degrees.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -796,23 +811,50 @@ void URArm::keep_alive_() {
...
+            std::stringstream err_string;
+
+            err_string << "rejecting move request : difference between starting trajectory position [(";
+            std::transform(std::make_move_iterator(waypoints.front().cbegin()),
+                           std::make_move_iterator(waypoints.front().cend()),
```
acmorrow, line 832: I'm surprised you need `make_move_iterator` here and below.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -1,12 +1,15 @@
 #include "ur_arm.hpp"
 #include "utils.hpp"
 
+#include <algorithm>
 #include <chrono>
 #include <cmath>
 #include <exception>
+#include <experimental/iterator>
```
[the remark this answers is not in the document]
npmenard: agreed i knew this was going to annoy you :) happy there is a boost alternative

src/viam/ur/module/ur_arm.cpp
npmenard: agreed and changed. Can you take a  look at get_joint_positions?

src/viam/ur/module/ur_arm.cpp
npmenard: nope

src/viam/ur/module/ur_arm.cpp
acmorrow: Agreed. I filed https://viam.atlassian.net/browse/RSDK-11547.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -562,10 +588,18 @@ void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
...
+    std::transform(std::make_move_iterator(curr_joint_position_rad.begin()),
+                   std::make_move_iterator(curr_joint_position_rad.end()),
+                   std::back_inserter(joint_position_degree),
+                   radians_to_degrees<double>);
+
+    return joint_position_degree;
 }
 
-std::vector<double> URArm::get_joint_positions_(const std::shared_lock<std::shared_mutex>&) {
+vector6d_t URArm::get_joint_positions_rad(const std::shared_lock<std::shared_mutex>&) {
```
acmorrow, line 602: nit: it is still private, throw a `_` on the end: `get_joint_positions_rad_`.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -562,10 +588,18 @@ void URArm::reconfigure(const Dependencies& deps, const ResourceConfig& cfg) {
...
-    return get_joint_positions_(rlock);
+    auto curr_joint_position_rad = get_joint_positions_rad(rlock);
+    std::vector<double> joint_position_degree;
+    joint_position_degree.reserve(curr_joint_position_rad.size());
+    std::transform(std::make_move_iterator(curr_joint_position_rad.begin()),
+                   std::make_move_iterator(curr_joint_position_rad.end()),
+                   std::back_inserter(joint_position_degree),
+                   radians_to_degrees<double>);
+
+    return joint_position_degree;
```
acmorrow, line 599: If you want to be "range-y":
```
    const auto degrees = radians | boost::adaptors::transformed([](const auto& r) { return radians_to_degrees(r); });
    return {begin(degrees), end(degrees)};
```

I think, since `vector6d` has random access iterators, the range does too, so this should also be O(1) distance calculation and I expect the vector constructor to leverage that to call `reserve` automatically before pushing the items.

src/viam/ur/module/ur_arm.cpp
acmorrow: Interesting. I'll try to take a look at that after you merge this.

acmorrow (approved):
lgtm, nits and suggestions only.

acmorrow (approved):
LGTM

PR #94, opened 2025-08-12
return optional to not throw on missing attributes

npmenard:


src/viam/ur/module/ur_arm.cpp
```diff
@@ -409,23 +408,18 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
     VIAM_SDK_LOG(info) << "appdir" << current_state_->appdir;
 
     // If the config contains `csv_output_path`, use that, otherwise,
-    // fall back to `VIAM_MOUDLE_DATA` as the output path, which must
+    // fall back to `VIAM_MODULE_DATA` as the output path, which must
     // be set.
     try {
-        current_state_->output_csv_dir_path = find_config_attribute<std::string>(cfg, "csv_output_path");
+        current_state_->output_csv_dir_path = find_config_attribute<std::string>(cfg, "csv_output_path").value();
+    } catch (std::invalid_argument& e) {
```
npmenard, line 415: would rather use value_or but it's argument is eagerly evaluated

src/viam/ur/module/ur_arm.cpp
```diff
@@ -111,22 +111,24 @@ void write_joint_data(const vector6d_t& jp, const vector6d_t& jv, std::ostream&
...
+    if (!find_config_attribute<double>(cfg, "speed_degs_per_sec")) {
+        throw std::invalid_argument("attribute `speed_degs_per_sec` is required");
+    }
+    if (!find_config_attribute<double>(cfg, "acceleration_degs_per_sec2")) {
+        throw std::invalid_argument("attribute `acceleration_degs_per_sec2` is required");
+    }
+
+    auto threshold = find_config_attribute<double>(cfg, "reject_move_request_threshold_deg");
+    constexpr double MIN_THRESHOLD = 0.0;
+    constexpr double MAX_THRESHOLD = 360.0;
```
acmorrow, line 126: I like these as `k_min_threshold` and `k_max_threshold`. C++ doesn't mandate all caps for constants (I think rust got this wrong), and I like this article for justification: https://accu.org/journals/overload/22/121/wakely_1923/

src/viam/ur/module/ur_arm.cpp
```diff
@@ -111,22 +111,24 @@ void write_joint_data(const vector6d_t& jp, const vector6d_t& jv, std::ostream&
...
+    if (threshold && (*threshold < MIN_THRESHOLD || *threshold > MAX_THRESHOLD)) {
+        std::stringstream sstream;
+        sstream << "attribute `reject_move_request_threshold_deg` should be between " << MIN_THRESHOLD << " and " << MAX_THRESHOLD
+                << " , it is : " << *threshold;
```
acmorrow, line 130: Maybe say "degrees".

src/viam/ur/module/utils.hpp
```diff
@@ -10,19 +12,18 @@ void configure_logger(const viam::sdk::ResourceConfig& cfg);
...
     auto key = cfg.attributes().find(attribute);
     if (key == cfg.attributes().end()) {
-        buffer << "required attribute `" << attribute << "` not found in configuration";
-        throw std::invalid_argument(buffer.str());
+        return std::nullopt;
     }
     const auto* const val = key->second.get<T>();
     if (!val) {
-        buffer << "required non-empty attribute `" << attribute << " could not be decoded";
+        buffer << "attribute `" << attribute << " could not be converted to the required type";
```
acmorrow, line 23: You could move the decl of `buffer` into the body of the if, since it isn't used higher up anymore.

src/viam/ur/module/utils.cpp
```diff
@@ -6,12 +6,8 @@
 #include <viam/sdk/resource/resource.hpp>
 
 void configure_logger(const viam::sdk::ResourceConfig& cfg) {
-    std::string level_str{};
-    try {
-        level_str = find_config_attribute<std::string>(cfg, "log_level");
-    } catch (...) {
-        level_str = "warn";
-    }
+    auto level_str = find_config_attribute<std::string>(cfg, "log_level").value_or("warn");
```
acmorrow, line 9: Nice. For some reason I thought `value_or` for `std::optional` wasn't in C++17, but it is!

src/viam/ur/module/ur_arm.cpp
acmorrow: I don't love deref'ing the empty optional with `value` and then catching the resulting exception. Maybe instead this:
```
current_state_->output_csv_dir_path = find_config_attribute<std::string>(cfg, "csv_output_path");
if (!current_state_->output_csv_dir_path) {
  // deal with VIAM_MODULE_DATA
}
```
That'll still throw if the argument doesn't convert.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -409,23 +408,18 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
-        if (cfg.attributes().count("csv_output_path") != 0) {
-            throw;
-        }
-
         auto* const viam_module_data = std::getenv("VIAM_MODULE_DATA");  // NOLINT: Yes, we know getenv isn't thread safe
         if (!viam_module_data) {
             throw std::runtime_error("required environment variable VIAM_MODULE_DATA unset");
         }
-        VIAM_SDK_LOG(info) << "VIAM_MODULE_DATA" << current_state_->output_csv_dir_path;
+        VIAM_SDK_LOG(warn) << "attribute `csv_output_path` set to VIAM_MODULE_DATA " << current_state_->output_csv_dir_path;
```
acmorrow, line 422: I actually think this is the common case, so, probably not worth a warning.

src/viam/ur/module/utils.hpp
```diff
@@ -10,19 +12,18 @@ void configure_logger(const viam::sdk::ResourceConfig& cfg);
 
 // helper function to extract an attribute value from its key within a ResourceConfig
 template <class T>
-T find_config_attribute(const viam::sdk::ResourceConfig& cfg, const std::string& attribute) {
+std::optional<T> find_config_attribute(const viam::sdk::ResourceConfig& cfg, const std::string& attribute) {
```
acmorrow, line 15: Regarding the case for the lack of `value_or` that doesn't eagerly evaluate, you could also have:

```
template<class T, class Callable>
T find_config_attribute(const Resourceconfig& config, const std::string& attribute, Callable&& c) {
    auto result = find_config_attribute(config, attribute);
    if (result) return *result;
    return std::forward<C>(c)();
}
```
Where you can pass in some lambda to compute the default. Then:

```
current_state_->output_csv_dir_path = find_config_attribute<std::string>(cfg, "csv_output_path", [] {
    // deal with VIAM_MODULE_DATA stuff
});
```
But I'll understand if you think that's taking it too far.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -389,15 +391,12 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
-        current_state_->reject_move_request_threshold_rad.emplace(degrees_to_radians(threshold));
-    }  // will go away once find_config_attribute returns an optional
-    catch (...) {
-        (void)0;
+    current_state_->host = find_config_attribute<std::string>(cfg, "host").value();
+    current_state_->speed.store(degrees_to_radians(find_config_attribute<double>(cfg, "speed_degs_per_sec").value()));
+    current_state_->acceleration.store(degrees_to_radians(find_config_attribute<double>(cfg, "acceleration_degs_per_sec2").value()));
+    auto threshold = find_config_attribute<double>(cfg, "reject_move_request_threshold_deg");
+    if (threshold) {
+        current_state_->reject_move_request_threshold_rad.emplace(degrees_to_radians(*threshold));
```
acmorrow, line 399: Or partially specialize `degrees_to_radians` for `std::optional<T>`. Just kidding, don't do that.

acmorrow (changes requested):
All comments are minor things, except maybe the constants so it is consistent with the rest of the file.

PR #95, opened 2025-08-12
[RSDK-11279] Refactor internal state management and background processing.

acmorrow:
First, apologies for the length and complexity here. It was hard to do this one incrementally since it was basically a rewrite of the entire "urcl" side of the driver.

Overall, this thing is more or less working, and it is far _far_ easier to understand the behavior of the system due to log messages like `URArm state transition from state independent(local) to state controlled due to event remote_mode_restored`. There are some rough edges around coming out of local mode that I'm working through with @JohnN193, but I don't expect they will challenge the overall architecture. There are also some small TODO items still to fix up, but, again, none of them are sufficient to prevent review from proceeding in parallel.

Other things that are overall missing:
- There are a number of areas that could use additional comments, and I plan to add those. Or have claude do it.
- Potentially, this new architecture is much more testable. However, without mocks for the various urcl components, I'd have no way to push the state machine around.

As usual, I will take a pass and leave PR notes to help guide the review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,13 +79,13 @@ endef
 # Targets for building AppImages
 appimage-arm64: export OUTPUT_NAME = universal-robots
 appimage-arm64: export ARCH = aarch64
-appimage-arm64: format build
+appimage-arm64: format-check build
 	$(call BUILD_APPIMAGE,$(OUTPUT_NAME),$(ARCH))
 	mv ./packaging/appimages/$(OUTPUT_NAME)-*-$(ARCH).AppImage* ./packaging/appimages/deploy/
 
 appimage-amd64: export OUTPUT_NAME = universal-robots
 appimage-amd64: export ARCH = x86_64
-appimage-amd64: format build
+appimage-amd64: format-check build
 	$(call BUILD_APPIMAGE,$(OUTPUT_NAME),$(ARCH))
 	mv ./packaging/appimages/$(OUTPUT_NAME)-*-$(ARCH).AppImage* ./packaging/appimages/deploy/
 
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -12,9 +12,11 @@
 #include <memory>
 #include <mutex>
 #include <optional>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <thread>
+#include <type_traits>
 #include <utility>
 #include <variant>
 
@@ -23,6 +25,7 @@
 #include <boost/range/adaptor/transformed.hpp>
 
 #include <ur_client_library/control/trajectory_point_interface.h>
+#include <ur_client_library/rtde/data_package.h>
 #include <ur_client_library/types.h>
 #include <ur_client_library/ur/dashboard_client.h>
 #include <ur_client_library/ur/ur_driver.h>
@@ -274,7 +277,7 @@ class URArm::state_ {
     struct state_independent_;
     friend struct state_independent_;
 
-    using state_variant = std::variant<state_disconnected_, state_controlled_, state_independent_>;
+    using state_variant_ = std::variant<state_disconnected_, state_controlled_, state_independent_>;
 
     struct event_connection_established_;
     struct event_connection_lost_;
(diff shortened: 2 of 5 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+
+URArm::state_::state_independent_::state_independent_(std::unique_ptr<arm_connection_> arm_conn, reason r)
+    : state_connected_(std::move(arm_conn)), reason_(r) {}
+
+std::string_view URArm::state_::state_independent_::name() {
+    using namespace std::literals::string_view_literals;
+    return "independent"sv;
+}
+
+std::string URArm::state_::state_independent_::describe() const {
```
acmorrow, line 1019: Here is where `describe` and such becomes useful: we can give the state a runtime description that includes whether we are estopped, local, or both.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+    // TODO: I'm not sure this actually makes sense: even if we are
+    // estopped, we still want high frequency service for the rtde
+    // interface.
+    if (estopped()) {
+        return k_estop_delay;
+    }
+    return k_noop_delay;
+}
+
+std::optional<URArm::state_::event_variant_> URArm::state_::state_independent_::upgrade_downgrade(state_&) {
```
acmorrow, line 1053: Expect changes here, @JohnN193 and I are still iterating on the exact recovery sequence for local mode.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
-    std::optional<move_request> move_request;
+bool URArm::state_::state_independent_::estopped() const {
+    return reason_ != reason::k_local_mode;
+}
+
+bool URArm::state_::state_independent_::local_mode() const {
+    return reason_ != reason::k_estopped;
+}
+
+URArm::state_::state_variant_ URArm::state_::state_independent_::handle_event(event_estop_cleared_) {
```
acmorrow, line 1193: Fairly boring event handlers from here. They just decide what the next state is given the current state, and instantiate a new state object.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+}
 
-    // variables specified by ResourceConfig and set through reconfigure
-    std::string host;
-    std::atomic<double> speed{0};
-    std::atomic<double> acceleration{0};
-    std::atomic<bool> estop{false};
-    std::atomic<bool> local_disconnect{false};
+template <typename T>
+void URArm::state_::emit_event_(T&& event) {
```
acmorrow, line 1352: The core of the state machine. Visits the current state, obtains its description and a description of the event it is about to handle. It then delegates to the current state's event handler for the current event, which will return the new state that we are going to become. We then get the description of that state, log the description, and return the new state. That newly returned state replaces our current state.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+    if (auto event = std::visit([this](auto& state) { return state.handle_move_request(*this); }, current_state_)) {
+        emit_event_(*std::move(event));
+    }
+}
+
+void URArm::state_::send_noop_() {
+    if (auto event = std::visit([](auto& state) { return state.send_noop(); }, current_state_)) {
+        emit_event_(*std::move(event));
+    }
+}
```
acmorrow, line 1400: These methods hide the state machine from the worker thread. They just visit the state and delegate to the per-state method of the same name. If the per-state method returns an event, we forward that to `emit_event_` to potentially advance the state machine to a different state.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+
+void URArm::state_::run_() {
```
acmorrow, line 1402: The high-frequency loop. We get the lock, use a condvar to wait either for the per-state timeout or the exit signal. If we wake up due to timeout, then it is time to drive the state machine. We call each activity in order. As we advance through the methods, the internal state may change method to method.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+            VIAM_SDK_LOG(warn) << "Unknown exception in worker thread";
+        }
+    }
+
+    VIAM_SDK_LOG(info) << "worker thread emitting disconnection event";
+    emit_event_(event_connection_lost_{});
+    VIAM_SDK_LOG(info) << "worker thread terminating";
+}
+
+void URArm::state_::trajectory_done_callback_(const control::TrajectoryResult trajectory_result) {
```
acmorrow, line 1429: Not much has changed here except that we can now swap the move request out under the lock but run the completion handlers without holding the lock, allowing the worker thread to get back to its job of communicating with the arm.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -331,9 +1499,9 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
 URArm::URArm(Model model, const Dependencies& deps, const ResourceConfig& cfg) : Arm(cfg.name()), model_(std::move(model)) {
     VIAM_SDK_LOG(info) << "URArm constructor called (model: " << model_.to_string() << ")";
     const std::unique_lock wlock(config_mutex_);
-    configure_(wlock, deps, cfg);
     // TODO: prevent multiple calls to configure_logger
     configure_logger(cfg);
+    configure_(wlock, deps, cfg);
```
acmorrow, line 1504: The other ordering meant you didn't get URCL logs during configuration.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -368,132 +1533,8 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
-        }
-        retry_count--;
-        std::this_thread::sleep_for(k_noop_delay);
-    }
-
-    // start background thread to continuously send no-ops and keep socket connection alive
-    VIAM_SDK_LOG(info) << "starting background_thread";
-    current_state_->keep_alive_thread = std::thread(&URArm::keep_alive_, this);
+    VIAM_SDK_LOG(info) << "URArm starting up";
+    current_state_ = state_::create(configured_model_type, cfg, ports_);
```
acmorrow, line 1537: All the work from here on that `configure` previously did is now handled by the state machine.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -508,48 +1549,6 @@ void URArm::check_configured_(const lock_type<std::shared_mutex>&) {
     }
 }
 
-void URArm::trajectory_done_cb_(const control::TrajectoryResult state) {
```
acmorrow, line 511: This mostly just moved, and is now part of `struct state_`, so it moved in with the other methods of that class.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -566,82 +1565,22 @@ std::vector<double> URArm::get_joint_positions(const ProtoStruct&) {
...
-    struct tm buf;
-    auto* ret = gmtime_r(&tt, &buf);
-    if (ret == nullptr) {
-        throw std::runtime_error("failed to convert time to iso8601");
-    }
-    stream << std::put_time(&buf, "%FT%T");
-    stream << "." << std::setw(6) << std::setfill('0') << delta_us.count() << "Z";
-
-    return stream.str();
+    const auto radians = current_state_->read_joint_positions();
```
acmorrow, line 1568: @npmenard - we are going to conflict a bit here, but we basically did the same thing.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -566,82 +1565,22 @@ std::vector<double> URArm::get_joint_positions(const ProtoStruct&) {
...
-
-    std::vector<double> to_ret;
-    for (const double joint_pos_rad : *current_state_->joints_position) {
-        const double joint_pos_deg = radians_to_degrees(joint_pos_rad);
-        to_ret.push_back(joint_pos_deg);
-    }
-    return to_ret;
-}
-
-std::string waypoints_filename(const std::string& path, const std::string& unix_time) {
```
acmorrow, line 588: These methods just moved up.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -566,82 +1565,22 @@ std::vector<double> URArm::get_joint_positions(const ProtoStruct&) {
...
+    const auto radians = current_state_->read_joint_positions();
+    const auto degrees = radians | boost::adaptors::transformed([](const auto& r) { return radians_to_degrees(r); });
+    return {begin(degrees), end(degrees)};
 }
 
 void URArm::move_to_joint_positions(const std::vector<double>& positions, const ProtoStruct&) {
     std::shared_lock rlock{config_mutex_};
     check_configured_(rlock);
 
-    if (current_state_->local_disconnect.load()) {
```
acmorrow, line 630: This check now happens state machine side.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -655,12 +1594,6 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
...
```
acmorrow, line 658: This check now happens state machine side.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -686,25 +1619,14 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
 pose URArm::get_end_position(const ProtoStruct&) {
     const std::shared_lock rlock{config_mutex_};
     check_configured_(rlock);
-
-    const std::lock_guard guard{current_state_->state_mutex};
```
acmorrow, line 690: This check now happens state machine side.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -686,25 +1619,14 @@ void URArm::move_through_joint_positions(const std::vector<std::vector<double>>&
...
-    return ur_vector_to_pose(*current_state_->tcp_state);
+    return ur_vector_to_pose(current_state_->read_tcp_pose());
 }
 
 bool URArm::is_moving() {
     const std::shared_lock rlock{config_mutex_};
     check_configured_(rlock);
-    return current_state_->trajectory_status.load() == TrajectoryStatus::k_running;
+    return current_state_->is_moving();
+    // return current_state_->trajectory_status.load() == TrajectoryStatus::k_running;
```
acmorrow, line 1629: stale, removed.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -755,50 +1677,35 @@ ProtoStruct URArm::do_command(const ProtoStruct& command) {
 
     ProtoStruct resp = ProtoStruct{};
 
+    // NOTE: Changes to these values will not be effective for any
```
acmorrow, line 1680: This isn't new, just documenting it.

acmorrow:
> I pressed send too soon. A few more notes:
> 
> * @npmenard I have not rebased this above [Add threshold to reject move request #93](https://github.com/viam-modules/universal-robots/pull/93) so don't be surprised that you don't see that here.
> * Besides fixing https://viam.atlassian.net/browse/RSDK-11279, this review also addresses the following tickets:
> * https://viam.atlassian.net/browse/RSDK-11290
> * https://viam.atlassian.net/browse/RSDK-11297
> * https://viam.atlassian.net/browse/RSDK-11298
> * https://viam.atlassian.net/browse/RSDK-11048
> * https://viam.atlassian.net/browse/RSDK-11444

Also https://viam.atlassian.net/browse/RSDK-11295

Makefile
```diff
@@ -79,13 +79,13 @@ endef
 # Targets for building AppImages
 appimage-arm64: export OUTPUT_NAME = universal-robots
 appimage-arm64: export ARCH = aarch64
-appimage-arm64: format build
+appimage-arm64: format-check build
```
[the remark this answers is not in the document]
JohnN193: fair, I had these added because I wanted to make sure format was being run at some point while testing. format-check should be good enough to catch that for me

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+    if (state.configured_model_type_ != actual_model_type) {
+        std::ostringstream buffer;
+        buffer << "configured model type `" << state.configured_model_type_ << "` does not match connected arm `" << actual_model_type
+               << "`";
+        throw std::runtime_error(buffer.str());
+    }
+
+    // If the arm is in remote control mode, we need to stop any
+    // running program, since it might not be ours.
+    if (arm_connection->dashboard->commandIsInRemoteControl()) {
```
JohnN193, line 726: kk here was what I found when investigating if we could remove the dashboard client/isInRemoteControl.

[isInRemoteControl](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/79672841a934e16a2ea84bc3eba6ca1720fd6cc5/src/ur/dashboard_client.cpp#L335C30-L341) sends a sendAndRecieve function. 

other apis such as commandPowerOff are similar in the [dashboard client](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/79672841a934e16a2ea84bc3eba6ca1720fd6cc5/src/ur/dashboard_client.cpp#L232-L236) but also have implementations in the [driver primary client](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/79672841a934e16a2ea84bc3eba6ca1720fd6cc5/src/primary/primary_client.cpp#L185-L202) using `driver->getPrimaryClient()->sendScript()`

so I believe we can implement commandIsInRemoteControl ourselves, which lets us remove the dashboard client entirely

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+}
+
+void URArm::state_::move_request::write_joint_data(vector6d_t& position, vector6d_t& velocity) {
+    ::write_joint_data(position, velocity, arm_joint_positions_stream, unix_time_iso8601(), arm_joint_positions_sample++);
+}
+
+URArm::state_::move_request::cancellation_request::cancellation_request() {}
+
+void URArm::state_::emit_event_(event_variant_&& event) {
+    std::visit([this](auto&& event) { this->emit_event_(std::forward<decltype(event)>(event)); }, std::move(event));
```
npmenard, line 1342: Does `std::forward<decltype(event)>(event)`  returns event_connection_established_, event_connection_established_,event_connection_lost_ etc... if is it evaluated at runtime?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+    struct event_remote_mode_restored_;
 
+    using event_variant_ = std::variant<event_connection_established_,
+                                        event_connection_lost_,
+                                        event_estop_detected_,
+                                        event_estop_cleared_,
+                                        event_local_mode_detected_,
+                                        event_remote_mode_restored_>;
+
+    struct state_disconnected_ {
```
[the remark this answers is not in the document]
npmenard: Since none of the state share a base classcan I declare a new state but without a `upgrade_downgrade` ? I imagine this would not compile

npmenard (approved):
Done with first pass. I only have comments pertaining to structure so i will approve it 
I like the way we can understand how to transition out of one state to just by looking which event are explicitly handled

JohnN193 (commented):
just a few callouts from testing so far. its incredibly easy to get the ur20

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +181,1281 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+    if (estopped() && arm_conn_->safety_status_bits->test(static_cast<size_t>(urtde::UrRtdeSafetyStatusBits::IS_NORMAL_MODE))) {
+        // TODO: This doesn't seem like entirely the right place for
+        // this. We end up in states where we are "paused" and we don't
+        // come out, and I think this is the thing that unsticks us. So
+        // this must need to happen in more cases than just this one.
+        try {
+            VIAM_SDK_LOG(info) << "While in independent state: releasing brakes since no longer estopped";
+            if (!arm_conn_->dashboard->commandBrakeRelease()) {
+                VIAM_SDK_LOG(warn) << "While in independent state, could not release brakes";
+                return std::nullopt;
```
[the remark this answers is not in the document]
acmorrow: Yes, sort of obvious now that you say it. I've filed https://viam.atlassian.net/browse/RSDK-11622 and added that ticket to the TODO above.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +181,1281 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+            return event_connection_lost_{};
         }
 
-        void complete_failure() {
-            complete_error("arm's current trajectory failed");
+        // TODO: I was experimenting here. Not sure if this is needed or not.
+        try {
+            VIAM_SDK_LOG(info) << "Arm has exited local control - cycling primary client";
+            arm_conn_->driver->stopPrimaryClientCommunication();
+            arm_conn_->driver->startPrimaryClientCommunication();
```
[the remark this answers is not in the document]
acmorrow: Same ticket as above and an added TODO here.

src/viam/ur/module/ur_arm.cpp
acmorrow: Since we have `auto&& event` as the parameter to the lambda, `event` is a forwarding reference. So we need to use `std::forward` on it. However, we don't actually have a typename handy to use. So we need to obtain that type by `decltype(event)`. It's clunky and C++20 (or 23?) has some things to make this less weird. Basically, this is just using forwarding references correctly.

Honestly thought this is overkill. I wrote these templates thinking that events might carry a lot of state and I was trying to keep the state machine as efficient as possible since it runs in the hot loop to talk to the arm. But most events are lightweight. I could probably rewrite all of this event passing machinery to use value semantics and `std::move` rather than && and `std::forward`. Let me know if you think that would be a sufficient gain in readability.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+
+    std::atomic<std::size_t> move_epoch_{0};
+    std::optional<move_request> move_request_;
+
+    struct ephemeral_ {
+        vector6d_t joint_positions;
+        vector6d_t joint_velocities;
+        vector6d_t tcp_state;
+    };
+    std::optional<struct ephemeral_> ephemeral_;
```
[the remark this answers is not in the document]
acmorrow: I filed https://viam.atlassian.net/browse/RSDK-11624 to track this idea.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -172,128 +184,1284 @@ void write_waypoints_to_csv(const std::string& filepath, const std::list<Eigen::
...
+    return state_independent_{std::move(arm_conn_), state_independent_::reason::k_estopped};
+}
+
+URArm::state_::state_variant_ URArm::state_::state_controlled_::handle_event(event_local_mode_detected_) {
+    return state_independent_{std::move(arm_conn_), state_independent_::reason::k_local_mode};
+}
+
+template <typename Event>
+URArm::state_::state_variant_ URArm::state_::state_controlled_::handle_event(Event) {
+    return std::move(*this);
```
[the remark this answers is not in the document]
acmorrow: I've filed https://viam.atlassian.net/browse/RSDK-11626 to remind me to revisit this later.

PR #99, opened 2025-08-15
[RSDK-11642] Make robot control frequency lower by default but config…

acmorrow:
…urable

### DIFF
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -260,8 +260,8 @@ class URArm::state_ {
                     std::string host,
                     std::string app_dir,
                     std::string csv_output_path,
-                    std::optional<double> robot_control_freq_hz,
                     std::optional<double> reject_move_request_threshold_rad,
+                    std::optional<double> robot_control_freq_hz,
                     const struct ports_& ports);
     ~state_();
 
--- a/src/viam/ur/module/ur_arm.cpp
+++ b/src/viam/ur/module/ur_arm.cpp
@@ -63,7 +63,6 @@ constexpr char k_output_recipe[] = "/src/control/rtde_output_recipe.txt";
 constexpr char k_input_recipe[] = "/src/control/rtde_input_recipe.txt";
 
 // constants for robot operation
-constexpr auto k_noop_delay = std::chrono::milliseconds(2);     // 2 millisecond, 500 Hz
 constexpr auto k_estop_delay = std::chrono::milliseconds(100);  // 100 millisecond, 10 Hz
 constexpr auto k_disconnect_delay = std::chrono::seconds(1);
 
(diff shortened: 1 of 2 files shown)


src/viam/ur/module/ur_arm.cpp
```diff
@@ -949,10 +978,6 @@ std::string URArm::state_::state_controlled_::describe() const {
     return std::string{name()};
 }
 
-std::chrono::milliseconds URArm::state_::state_controlled_::get_timeout() const {
-    return k_noop_delay;
```
JohnN193, line 953: actually, can we go ahead and delete k_noop_delay?

acmorrow: Fixed, below

src/viam/ur/module/ur_arm.cpp
```diff
@@ -1513,13 +1517,45 @@ void URArm::state_::send_noop_() {
...
+            last_sampling_point = wait_start;
+            accumulator.emplace();
+        }
+
+        if (accumulator) {
+            auto wait_end = std::chrono::steady_clock::now();
+            (*accumulator)(std::chrono::duration_cast<std::chrono::duration<double, std::milli>>(wait_end - wait_start).count());
+            if (bacc::count(*accumulator) == k_num_samples) {
+                const auto accumulated = std::exchange(accumulator, std::nullopt);
+                VIAM_SDK_LOG(info) << "URArm worker thread mean wait between control cycles is " << bacc::mean(*accumulated)
```
npmenard, line 1554: mean is not super relevant here IMO median is better

src/viam/ur/module/ur_arm.cpp
```diff
@@ -1513,13 +1517,45 @@ void URArm::state_::send_noop_() {
...
+    // Periodically, collect a limited number of samples of the
+    // duration of our wait latency on the condition variable. We
+    // expect this to fairly well respect the configured
+    // `robot_control_freq_hz`. Report this in the log so that we can
+    // know how well we are doing staying a reliable consumer of arm
+    // data. The logging also serves as a proof of forward progress
+    // for the worker thread.
+    constexpr std::size_t k_num_samples = 100;
+    constexpr auto k_sampling_interval = std::chrono::minutes(5);
+    auto last_sampling_point = std::chrono::steady_clock::now();
```
npmenard, line 1530: not against per se but not sure how useful it would be since these are snapshots of [0.2;1]s ([500,100]) I would leave it tough just to get an initial sense of perfs. For it to be very useful (like following how the median shifts while sanding does the heavy calcs) we would need to invest a lot more time

npmenard (approved):
LGTM

src/viam/ur/module/ur_arm.cpp
acmorrow: Yes, agreed, but at least this 1) gives us an ongoing liveness indication, which I think is valuable, and 2) provides some information we didn't have before about how well we are hitting our obligations. We can iterate on it. I would love it if we had a more general program for giving components a way to surface statistics on their behavior!

PR #104, opened 2025-09-16
[RSDK-11923] Do connection establishment out of disconnected with a background thread

acmorrow:
This allows calls to `is_moving` to be answered while connection is in progress, since we don't block with the state mutex in `disconnected::upgrade_downgrade` but instead repeatedly invoke it after re-acquiring the mutex.

Note that this is a partial solution, because we still have blocking connect calls in the `state_independent_`. The next step for this ticket, in a follow-up PR, will be to transition through the disconnected state when the system traverses a local-to- remote or remote-to-local edge, which will centralize all connection establishment onto this concurrent scheme. My hope is that this will allow us to significantly simplify the logic in the independent state, ensuring that we get accurate answers when querying the dashboard for the local/remote state.

### DIFF
--- a/src/viam/ur/module/ur_arm_state.cpp
+++ b/src/viam/ur/module/ur_arm_state.cpp
@@ -2,6 +2,7 @@
 
 #include <filesystem>
 #include <stdexcept>
+#include <thread>
 
 #include <boost/accumulators/accumulators.hpp>
 #include <boost/accumulators/statistics/median.hpp>
@@ -130,30 +131,38 @@ std::unique_ptr<URArm::state_> URArm::state_::create(std::string configured_mode
                                           std::move(csv_output_path),
                                           std::move(threshold),
                                           std::move(frequency),
-
                                           ports);
 
     state->set_speed(degrees_to_radians(find_config_attribute<double>(config, "speed_degs_per_sec").value()));
     state->set_acceleration(degrees_to_radians(find_config_attribute<double>(config, "acceleration_degs_per_sec2").value()));
 
+    // Hold the mutex while we start the worker thread. It will not be
+    // able to advance, but will be ready to take over work as soon as
+    // we release the lock, minimizing the delay between establishing
+    // a connection and allowing the worker thread to begin meeting
+    // its service obligations.
     const std::lock_guard lock(state->mutex_);
     state->worker_thread_ = std::thread{&state_::run_, state.get()};
 
     // Attempt to manually drive the state machine out of the
     // disconnected state. For now, just try a few times with the
     // natural recovery cycle.
-    for (size_t i = 0; i != 5; ++i) {
+    size_t connection_failures = 0;
+    constexpr size_t max_connection_failures = 5;
+    while (std::holds_alternative<state_disconnected_>(state->current_state_)) {
         try {
             state->upgrade_downgrade_();
-            return state;
         } catch (const std::exception& xcp) {
-            VIAM_SDK_LOG(warn) << "Failed to establish working connection to arm after " << (i + 1) << " attempts: `" << xcp.what() << "`";
-            const auto timeout = state->get_timeout_();
-            VIAM_SDK_LOG(warn) << "Retrying after " << timeout.count() << " milliseconds";
-            std::this_thread::sleep_for(timeout);
+            VIAM_SDK_LOG(warn) << "Failed to establish working connection to arm after " << ++connection_failures << " attempts: `"
+                               << xcp.what() << "`";
+            if (connection_failures >= max_connection_failures) {
+                throw;
+            }
+            VIAM_SDK_LOG(warn) << "Retrying after " << state->get_timeout_().count() << " milliseconds";
         }
+        std::this_thread::sleep_for(state->get_timeout_());
     }
-    state->upgrade_downgrade_();
+
     return state;
 }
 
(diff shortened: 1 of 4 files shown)


src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -130,30 +131,38 @@ std::unique_ptr<URArm::state_> URArm::state_::create(std::string configured_mode
                                           std::move(csv_output_path),
                                           std::move(threshold),
                                           std::move(frequency),
-
                                           ports);
 
     state->set_speed(degrees_to_radians(find_config_attribute<double>(config, "speed_degs_per_sec").value()));
     state->set_acceleration(degrees_to_radians(find_config_attribute<double>(config, "acceleration_degs_per_sec2").value()));
 
+    // Hold the mutex while we start the worker thread. It will not be
```
acmorrow, line 139: This was always true but I realized it wasn't obvious at all, so I left a comment explaining how it works.

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -130,30 +131,38 @@ std::unique_ptr<URArm::state_> URArm::state_::create(std::string configured_mode
...
+    // a connection and allowing the worker thread to begin meeting
+    // its service obligations.
     const std::lock_guard lock(state->mutex_);
     state->worker_thread_ = std::thread{&state_::run_, state.get()};
 
     // Attempt to manually drive the state machine out of the
     // disconnected state. For now, just try a few times with the
     // natural recovery cycle.
-    for (size_t i = 0; i != 5; ++i) {
+    size_t connection_failures = 0;
```
acmorrow, line 150: This loop needed a refactor because calls to `upgrade_downgrade_` now may be either initiating a connection or polling for completion, so just calling it 5 times doesn't work anymore. Instead, we just call it repeatedly until we either get 5 failures or a success. Effectively, that's the same observable behavior.

src/viam/ur/module/ur_arm_state_connected.cpp
```diff
@@ -19,7 +19,7 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_connected_::se
 std::optional<URArm::state_::event_variant_> URArm::state_::state_connected_::recv_arm_data(state_& state) {
     const auto prior_robot_status_bits = std::exchange(arm_conn_->robot_status_bits, std::nullopt);
     const auto prior_safety_status_bits = std::exchange(arm_conn_->safety_status_bits, std::nullopt);
-    // arm_conn_->data_package
+
```
acmorrow, line 22: This was stray

src/viam/ur/module/ur_arm_state_disconnected.cpp
```diff
@@ -12,23 +12,53 @@ std::string URArm::state_::state_disconnected_::describe() const {
...
+        // event. If the future isn't ready, there is no event to
+        // return, and we will come back to poll again.
+        switch (pending_connection->wait_for(std::chrono::seconds(0))) {
+            case std::future_status::ready: {
+                return event_connection_established_{std::exchange(pending_connection, std::nullopt)->get()};
+            }
+            case std::future_status::timeout: {
+                break;
+            }
+            case std::future_status::deferred: {
```
JohnN193, line 40: are these the only cases? do we not need a default?

src/viam/ur/module/ur_arm_state_disconnected.cpp
```diff
@@ -12,23 +12,53 @@ std::string URArm::state_::state_disconnected_::describe() const {
 }
 
 std::chrono::milliseconds URArm::state_::state_disconnected_::get_timeout() const {
-    return std::chrono::seconds(1);
+    // If we have a pending connection, we are polling for completion,
+    // so increase the sampling rate. Otherwise we want to wait a more
+    // reasonable length of time before attempting to connect again.
+    return pending_connection ? std::chrono::milliseconds(5) : std::chrono::seconds(1);
```
JohnN193, line 18: while the arm is turned off, do we have an idea of how long we are in the pending_connection state? 

this makes sense because we only try connecting to the arm at most once per second, but I don't remember how long connection attempts would take(seemingly long because we are doing this ticket)

src/viam/ur/module/ur_arm_state_disconnected.cpp
```diff
@@ -116,7 +146,7 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_disconnected_:
     }
 
     VIAM_SDK_LOG(info) << "disconnected: attempting recovery: recovery appears to have been successful; transitioning to independent mode";
-    return event_connection_established_{std::move(arm_connection)};
+    return arm_connection;
```
JohnN193, line 149: so we don't need a move to pass the unique ptr? or does the `pending_connection.emplace` bit handle all of that?

JohnN193 (approved):
lgtm, just some nit questions

src/viam/ur/module/ur_arm_state_disconnected.cpp
acmorrow: It isn't so much that they are long, exactly. More that we were holding the state machine lock while we were doing it, which made it impossible to answer queries about our state (which, naturally, must also take the lock so they can observe stable values as updated by the running state machine).

src/viam/ur/module/ur_arm_state_disconnected.cpp
acmorrow: These cases are exhaustive so no default is needed. If std::future_status ever grew a new value, the compiler or linter would almost certainly whine at us. In fact, I'd have preferred to not write one for `deferred` at all, but if I don't, I get a tidy warning:
```
/src/src/viam/ur/module/ur_arm_state_disconnected.cpp:33:17: error: enumeration value 'deferred' not handled in switch [-Werror,-Wswitch]
   33 |         switch (pending_connection->wait_for(std::chrono::seconds(0))) {
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

src/viam/ur/module/ur_arm_state_disconnected.cpp
acmorrow: We don't need a move here because we are returning a unique_ptr, and returning like that automatically moves. The old code was returning an `event_connection_established_` and we needed to move the `arm_connection` into that object.

PR #110, opened 2025-09-25
Arm status

JohnN193:


JohnN193:
other thought I had was parsing the describe into simpler statements, rather than posting the full independent state. I believe this is more in line with that Esha recommended, but I would want more feedback first

```
std::string parse_describe(const std::string describe) {
    if (describe == "controlled") {
        return "arm is ready to move";
    } else if (describe == "disconnected") {
        return "arm appears to be offline";
    }
    std::ostringstream buffer;
    bool is_local = false;
    if (describe.find("local") != std::string::npos) {
        is_local = true;
        buffer << "the arm is in local mode";
    }
    if (describe.find("stop") != std::string::npos) {
        if (describe.find("NORMAL") != std::string::npos) {
            return buffer.str();
        }
        if (is_local) {
            buffer << " and ";
        }
        if (describe.find("PROTECTIVE") != std::string::npos) {
            buffer << "arm is protective stopped";
        } else {
            buffer << "arm is stopped";
        }
    }
    return buffer.str();
}
```

src/viam/ur/module/ur_arm.cpp
```diff
@@ -422,9 +422,13 @@ ProtoStruct URArm::do_command(const ProtoStruct& command) {
     constexpr char k_get_tcp_forces_base_key[] = "get_tcp_forces_base";
     constexpr char k_get_tcp_forces_tool_key[] = "get_tcp_forces_tool";
     constexpr char k_clear_pstop[] = "clear_pstop";
+    constexpr char k_arm_ready[] = "arm_ready";
+    constexpr char k_arm_status[] = "arm_status";
```
acmorrow, line 426: - These are, effectively, method names, so I think they should start with a verb, like `get_` or `is_`.
- We are an arm so I think (hope?) we don't need that in the name
- Maybe `is_[in_]controllable_state` and `get_status_description` or `get_state_description`?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -422,9 +422,13 @@ ProtoStruct URArm::do_command(const ProtoStruct& command) {
...
 
     // Cache TCP state to ensure atomic read of pose and forces from same timestamp
     std::optional<decltype(current_state_->read_tcp_state_snapshot())> cached_tcp_state;
+    // Cache arm state description
+    auto description = current_state_->describe();
```
acmorrow, line 431: Agree on getting the value outside the loop, but you could use the same `optional` trick I do above to avoid getting it if you don't or won't need it.

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -57,6 +57,7 @@ class URArm::state_ {
                                            std::ofstream arm_joint_positions_stream);
 
     bool is_moving() const;
+    std::string describe() const;
```
acmorrow, line 60: Keep this method, even if we need to add others. It is likely to prove useful in the future.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -464,6 +468,14 @@ ProtoStruct URArm::do_command(const ProtoStruct& command) {
         } else if (kv.first == k_clear_pstop) {
             current_state_->clear_pstop();
             resp.emplace(k_clear_pstop, "protective stop cleared");
+        } else if (kv.first == k_arm_ready) {
+            if (description == "controlled") {
+                resp.emplace(k_arm_ready, true);
+            } else {
+                resp.emplace(k_arm_ready, false);
```
acmorrow, line 475: I think I'd prefer not to parse the state description, but to have a method on the actual state class that obtains this bool. Partially because descriptions could change and then this method would silently break, and also because the state machine might later evolve other states of which more than one represented a controlled state.

Making it a method on state raises a little complication, because we want to get a consistent answer to potentially two queries, one for whether the arm is in a controllable state, and the other for the log-aligned description of the current state.

Adding an `is_current_state_controlled` to `class state_` is easy enough, and you already have `describe`.

But there isn't (currently) a way to get both of those at the same time without releasing the lock between the queries. If you called both `is_current_staet_controlled` and `describe` you might get inconsistent results if a state transition happened in between.

We can either decide that that is an issue not worth worrying about because it is unlikely to happen in practice, in which case just call the new method in one place and `describe` in the other, and then you don't need to do the caching.

Or, we can decide that we do care about making sure the answer is consistent if you ask for both. That'll require caching, and a single method to call. What should that method look like?

I think maybe the easiest answer is to make `is_current_state_controlled` take an optional out parameter to capture the description:

```
bool URArm::state_::is_current_state_controlled(std::string* description = nullptr);
```

The idea is that you may only care about whether the state is controlled or not. But you might also be in a position where if the arm is, say, not controlled, you want to report the description of the state as well. That function would arguably cover that, and do what you need here.

Then in this function, you can just make a little struct local to the function:

```
struct controlled_info {
    bool controlled;
    std::string description;
};
std::optional<controlled_info> cached_controlled_info;
```

src/viam/ur/module/ur_arm.cpp
acmorrow, line 426: Don't forget to add these new commands to the documentation.

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -414,10 +414,23 @@ std::string URArm::state_::describe_() const {
...
 std::string URArm::state_::describe_state_(const state_variant_& state) {
     return std::visit([](auto& state) { return state.describe(); }, state);
 }
 
+bool URArm::state_::is_current_state_controlled(std::string* description /*= nullptr*/) const {
+    const std::lock_guard lock{mutex_};
+    if (description != nullptr) {
+        *description = describe_();
+    }
+    return current_state_.index() == 1;
```
JohnN193, line 431: I couldn't figure out a way to put `state_controlled_` here, was there a quick way using `state_variant` that I'm missing or should I include something else? Or is this sufficient for now?

PR #111, opened 2025-09-25
Resume trajectory

JohnN193:


src/viam/ur/module/ur_arm_state_controlled.cpp
```diff
@@ -62,14 +62,16 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::h
         return std::nullopt;
     }
 
-    if (!state.move_request_->samples.empty() && !state.move_request_->cancellation_request) {
-        // We have a move request, it has samples, and there is no pending cancel for that move. Issue the move.
+    if ((state.move_request_->trajectory_start.time_since_epoch().count() == 0) && !state.move_request_->cancellation_request) {
```
JohnN193, line 65: switched to using time here instead of emptying the samples, so that when a failure occurs I can pull the samples from the move request.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -529,7 +529,7 @@ void URArm::resume_trajectory_(std::shared_lock<std::shared_mutex> config_rlock)
     }
     auto failed_trajectory = current_state_->get_failed_trajectory();
 
-    if (failed_trajectory.size() == 0) {
+    if (failed_trajectory.empty()) {
```
nfranczak, line 532: [q]: what is the difference between these two?

JohnN193: the linter yells at me if I do the first one

JohnN193:
closing for now until we want to think about this again

PR #114, opened 2025-10-02
add ur7e

JohnN193:
adds a ur7e model. note this pr assumes

1. that the kinematics of the ur7e are the same as the ur5e
2. that the ur7e will report its model as UR7 and not UR5


acmorrow (approved):
LGTM

PR #197, opened 2026-03-30
[RSDK-12645] Bump urcl

JohnN193:
https://viam.atlassian.net/browse/RSDK-12645

  Bump Universal Robots Client Library from 2.2.0 to 2.9.0                              
                                                                                        
  This PR updates the urcl dependency and adapts the module to its changed API.         
                                                                                        
  API changes (urcl 2.x)                                                                
                                                                                        
  - getDataPackage() no longer returns a unique_ptr<DataPackage> — it now takes a       
  pre-allocated DataPackage& and returns bool. Updated all call sites in
  state_connected_ and state_disconnected_ accordingly. On failure, data_package is left
   unmodified so the last good packet is retained.          
  - DashboardClient moved into the urcl:: namespace. Updated construction and type
  annotations in arm_connection_ and state_disconnected_.                               
   
  Behavioral changes                                                                    

  - Updates urscript controls file to handle trajectory cancellation correctly(no more pstops on cancel)                                                         
  - Removed dashboard socket-state checks in state_controlled_ and state_independent_.  
  These checked dashboard->getState() == Connected and dropped the connection if not. now we just rely on errors from urcl to detect the change. since we already kicked into disconnected when we would detect local mode its really not that different
  - The fix for silent network loss (wifi-drop causing RTDEClient::~RTDEClient() to hang
   indefinitely)

### DIFF
--- a/src/viam/ur/module/ur_arm_state.hpp
+++ b/src/viam/ur/module/ur_arm_state.hpp
@@ -182,7 +182,7 @@ class URArm::state_ {
 
         ~arm_connection_();
 
-        std::unique_ptr<urcl::DashboardClient> dashboard;
+        std::unique_ptr<DashboardClient> dashboard;
         std::unique_ptr<UrDriver> driver;
         std::unique_ptr<rtde_interface::DataPackage> data_package;
         std::optional<std::bitset<k_num_robot_status_bits>> robot_status_bits;
--- a/src/viam/ur/module/ur_arm_state.hpp
+++ b/src/viam/ur/module/ur_arm_state.hpp
@@ -182,7 +182,7 @@ class URArm::state_ {
 
         ~arm_connection_();
 
-        std::unique_ptr<DashboardClient> dashboard;
+        std::unique_ptr<urcl::DashboardClient> dashboard;
         std::unique_ptr<UrDriver> driver;
         std::unique_ptr<rtde_interface::DataPackage> data_package;
         std::optional<std::bitset<k_num_robot_status_bits>> robot_status_bits;
--- a/src/viam/ur/module/ur_arm_state_controlled.cpp
+++ b/src/viam/ur/module/ur_arm_state_controlled.cpp
@@ -33,11 +33,6 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::u
         return event_stop_detected_{};
     }
 
-    if (arm_conn_->dashboard->getState() != urcl::comm::SocketState::Connected) {
-        VIAM_SDK_LOG(warn) << "While in state " << describe() << ", dashboard client is disconnected; dropping connection";
-        return event_connection_lost_::dashboard_communication_failure();
-    }
-
     // If we get anything but a positive answer from the dashboard
     // that we are in remote control, assume that we need to
     // completely re-create our connection, since sockets aren't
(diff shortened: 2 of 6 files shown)


npmenard (commented):
LGTM but have andrew double check it
Did they merge your fix in the urcl library?

npmenard (approved):

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -182,7 +182,7 @@ class URArm::state_ {
 
         ~arm_connection_();
 
-        std::unique_ptr<DashboardClient> dashboard;
+        std::unique_ptr<urcl::DashboardClient> dashboard;
```
acmorrow: I'm surprised you needed this. In `ur_arm.hpp`, which we include here, we do `using namespace urcl` (perhaps that is ill-advised, but, it is what is happening).

src/viam/ur/module/ur_arm_state_disconnected.cpp
```diff
@@ -149,7 +149,8 @@ std::unique_ptr<URArm::state_::arm_connection_> URArm::state_::state_disconnecte
     arm_connection->driver->startRTDECommunication();
 
     VIAM_SDK_LOG(debug) << "While in state " << describe() << ", attempting to reconnect: attempting to read a data package from the arm";
-    if (!arm_connection->driver->getDataPackage()) {
+    arm_connection->data_package = std::make_unique<urcl::rtde_interface::DataPackage>(arm_connection->driver->getRTDEOutputRecipe());
```
acmorrow, line 152: Interesting that they didn't create a `UrDriver::allocateDataPackage` or similar. This seems sort of clunky that we need to do it this way. I poked around and didn't see anything though, and their examples seem to do it this way too.

src/viam/ur/module/ur_arm_state_controlled.cpp
```diff
@@ -33,11 +33,6 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::u
         return event_stop_detected_{};
     }
 
-    if (arm_conn_->dashboard->getState() != urcl::comm::SocketState::Connected) {
```
acmorrow, line 36: Interesting. So, in the version of URCL we currently use, `DashboardClient` isa `TCPSocket` and that is where we get the ability to call `getState`. In the new version used here, `DashBoardClient` no longer isa `TCPSocket` so we can't make that call.

It looks like this was done in furtherance of future (or now available?) support for PolyScope X.

I took a look at the new APIs and I don't see any way to replace what this was doing, so I don't see that we have much alternative except to stop doing this.

It makes me a little nervous though, in that the state machine has been working pretty much flawlessly for quite some time and I'm reluctant to change it.

acmorrow (approved):
LGTM, though I don't expect you really need the `urcl::` qualification. What's the best argument for why dropping the explicit dashboard client state checks is safe?

src/viam/ur/module/ur_arm_state_controlled.cpp
JohnN193: yeah theres currently no way to access it, and seemingly  the only api that checks the socket state is `dashboard->connect()`. 

looking at their code some more, the socket state only changes when a `TCPSocket::read()` call fails or the socket is explicitly closed, so I'm not actually sure if the `getState() `check ever did anything in the first place. The majority of their apis would make `sendAndReceive` calls that should have errored somewhere and triggered the event in our state machine.

PR #200, opened 2026-04-03
remove old plotting

JohnN193:
we are slowly moving from csv to json files, we have other plotting scripts, and we get spammed by prs trying to fix this. lets finally remove it


acmorrow (approved):
LGTM

PR #204, opened 2026-04-22
Add realtime trajectory logging

npmenard:


src/viam/ur/module/trajectory_logger.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+void RealtimeTrajectoryLogger::append_realtime_sample(const std::string& timestamp_iso,
+                                                      const ephemeral_data& data,
+                                                      std::optional<uint32_t> robot_status_bits,
+                                                      std::optional<uint32_t> safety_status_bits) {
+    Json::Value sample;
+    sample["timestamp"] = timestamp_iso;
+    sample["positions_rad"] = vector6d_to_json(data.joint_positions);
+    sample["velocities_rad_per_sec"] = vector6d_to_json(data.joint_velocities);
```
acmorrow, line 90: What about actual accel, as `accelerations_rad_per_sec2`? Below you have target.

src/viam/ur/module/trajectory_logger.cpp
```diff
@@ -0,0 +1,158 @@
...
+        samples);
+    root_["planned_trajectory"] = arr;
+}
+
+void RealtimeTrajectoryLogger::append_realtime_sample(const std::string& timestamp_iso,
+                                                      const ephemeral_data& data,
+                                                      std::optional<uint32_t> robot_status_bits,
+                                                      std::optional<uint32_t> safety_status_bits) {
+    Json::Value sample;
+    sample["timestamp"] = timestamp_iso;
```
acmorrow: I think this will mean times are recorded as strings? Is that intentional?

src/viam/ur/module/trajectory_logger.cpp
```diff
@@ -0,0 +1,158 @@
...
+        std::cerr << "RealtimeTrajectoryLogger could not open file: " << output_path_ << '\n';
+        return;
+    }
+    out << json_str;
+    out.close();
+}
+
+std::string RealtimeTrajectoryLogger::realtime_trajectory_filename(const std::string& path,
+                                                                   const std::string& resource_name,
+                                                                   const std::string& unix_time) {
```
acmorrow, line 133: Not for this review unless you are interested, but this using the "unix time" as the move ID is pretty weak. We should just generate a UUID or increment a counter. These times just muddy the picture since they only have anything to do with when the motion started.

src/viam/ur/module/trajectory_logger.hpp
```diff
@@ -0,0 +1,69 @@
...
+class RealtimeTrajectoryLogger {
+   public:
+    RealtimeTrajectoryLogger(const std::filesystem::path& telemetry_path,
+                             const std::string& timestamp,
+                             const std::string& robot_model,
+                             const std::string& resource_name);
+    ~RealtimeTrajectoryLogger();
+
+    RealtimeTrajectoryLogger(RealtimeTrajectoryLogger&& other) noexcept;
+    RealtimeTrajectoryLogger& operator=(RealtimeTrajectoryLogger&& other) noexcept;
```
acmorrow: Where does the need for moveability of this type come in? Might be better to just allocate it and let it move around via unique_ptr?

src/viam/ur/module/trajectory_logger.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <viam/trajex/totg/waypoint_accumulator.hpp>
+
+#include "ur_arm.hpp"
+
+using namespace urcl;
+
+class RealtimeTrajectoryLogger {
+   public:
+    RealtimeTrajectoryLogger(const std::filesystem::path& telemetry_path,
+                             const std::string& timestamp,
```
acmorrow, line 19: Is this `timestamp` the `unix_time` we use to name files? If so, I'd keep that naming, even though it is odd, so that when we move to some other scheme to identify these files we can fix them all in one pass with less ambiguity.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -793,13 +719,15 @@ void URArm::move_tool_space_(std::shared_lock<std::shared_mutex> config_rlock, p
...
-    // For pose-space moves, we don't log joint data since we only have the target pose
-    auto trajectory_completion_future = [&, config_rlock = std::move(our_config_rlock)]() mutable {
-        return current_state_->enqueue_move_request(current_move_epoch, std::nullopt, std::move(async_cancellation_monitor), ps);
+    RealtimeTrajectoryLogger logger(telemetry_path, unix_time, model_.to_string(), current_state_->resource_name());
+    logger.set_velocity_limits(current_state_->get_velocity_limits());
+    logger.set_acceleration_limits(current_state_->get_acceleration_limits());
+
+    auto trajectory_completion_future = [&, config_rlock = std::move(our_config_rlock), logger = std::move(logger)]() mutable {
+        return current_state_->enqueue_move_request(
+            current_move_epoch, std::optional<RealtimeTrajectoryLogger>{std::move(logger)}, std::move(async_cancellation_monitor), ps);
```
acmorrow: Using unique_ptr might let you drop this optional since unique_ptr has its own optionality, and then you wouldn't need any of the move semantics on the logger (you could explicitly make it non-moveable by deleting the move ops).

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -598,9 +573,29 @@ void URArm::state_::move_request::cancel_error(std::string_view message) {
...
 }
 
-void URArm::state_::move_request::write_joint_data(vector6d_t& position, vector6d_t& velocity) {
-    if (arm_joint_positions_stream) {
-        ::write_joint_data(position, velocity, *arm_joint_positions_stream, unix_time_iso8601(), arm_joint_positions_sample++);
+void URArm::state_::move_request::write_realtime_sample(const struct URArm::state_::ephemeral_& data,
+                                                        std::optional<uint32_t> robot_status_bits,
+                                                        std::optional<uint32_t> safety_status_bits) {
+    if (trajectory_logger) {
+        trajectory_logger->append_realtime_sample(unix_time_iso8601(),
```
acmorrow: The call to `unix_time_iso8601` isn't cheap. I'd suggest, as I did elsewhere, not capturing this as a string but instead as a number.

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -435,6 +448,16 @@ class URArm::state_ {
         vector6d_t joint_velocities;
         vector6d_t tcp_state;
         vector6d_t tcp_forces;
+        vector6d_t target_joint_positions;
```
acmorrow: I think I'd prefer that there only be one definition of the `ephemeral_` data, and that the logger and the state machine shared it. There is nothing inherently protected about this structure that requires it to be private to URArm::state, or to the logger, or even nested inside a class. You could probably just promote it to whatever level / position allows both state and your logger to use the same declaration.

src/viam/ur/module/ur_arm_state_connected.cpp
```diff
@@ -70,6 +70,16 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_connected_::re
     static const std::string k_joints_velocity_key = "actual_qd";
     static const std::string k_tcp_key = "actual_TCP_pose";
     static const std::string k_tcp_force_key = "actual_TCP_force";
+    static const std::string k_target_q_key = "target_q";
```
acmorrow, line 80: Can you alphabetize this list starting at line 69

src/viam/ur/module/ur_arm_state_controlled.cpp
```diff
@@ -134,7 +134,14 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::h
 
                 } else {
                     // TODO: is it assured that we have positions/velocities here?
-                    state.move_request_->write_joint_data(state.ephemeral_->joint_positions, state.ephemeral_->joint_velocities);
+                    state.move_request_->write_realtime_sample(
+                        *state.ephemeral_,
+                        arm_conn_->robot_status_bits
```
acmorrow, line 139: Monadics on optional can't come soon enough.

src/control/rtde_output_recipe.txt
```diff
@@ -25,4 +25,14 @@ safety_mode
 robot_status_bits
 safety_status_bits
 actual_current
-tcp_offset
\ No newline at end of file
+tcp_offset
```
acmorrow: Maybe sort-lines on this file

src/viam/ur/module/ur_arm_state_independent.cpp
```diff
@@ -208,24 +208,62 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_independent_::
...
-        }());
+            return std::nullopt;
+        }
+
+        // Cooldown expired — fall through to destroy the move request.
+        VIAM_SDK_LOG(debug) << "post-stop telemetry recording complete";
     }
+
+    post_stop_recording_deadline_.reset();
+    std::exchange(state.move_request_, {})->complete_error([this]() -> std::string {
```
acmorrow: I think, if I'm reading this right, that this will result in a 1 second delay before the completion of the move request. In other words, the client won't learn of the move resulting in an error for one second after it happens. Is that intended?

acmorrow (changes requested):
This looks great. High level, my concerns, in order of importance are:

- I'm concerned about delaying the response back to the client about the fate of the trajectory due to a stop. We might need another scheme.
- I'd like to eliminate the duplication of the ephemeral state so the two participants can share a definition.
- I think the times should be recorded as numbers, not formatted strings.
- I think you can supplant both move semantics for RealtimeTrajectoryLogger and the optionality of having one by using std::unique_ptr.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -870,8 +800,7 @@ void URArm::move_joint_space_(std::shared_lock<std::shared_mutex> config_rlock,
             viam::trajex::totg::waypoint_accumulator accumulator{*current_pos};
             accumulator.add_waypoints(waypoints);
 
-            write_waypoints_to_csv(waypoints_filename(current_state_->telemetry_output_path(), current_state_->resource_name(), unix_time),
-                                   accumulator);
+            captured_waypoints = accumulator;
```
acmorrow, line 803: This violates the spirit, just a bit, of the planner and the receiver type and all that, but tbh I don't care that much. The whole planner concept goes away once the legacy trajectory generator goes away, so, we will just clean it up when we get to that.

src/viam/ur/module/trajectory_logger.cpp
npmenard: Yes but I don't have a strong preference you think u64 timestamps are better?

src/viam/ur/module/trajectory_logger.cpp
npmenard: not exported, closest is actual_current

src/viam/ur/module/ur_arm_state.hpp
npmenard: done

src/viam/ur/module/ur_arm_state_connected.cpp
npmenard: done

src/viam/ur/module/ur_arm_state_independent.cpp
npmenard: updated

src/viam/ur/module/ur_arm.hpp
```diff
@@ -46,6 +46,26 @@ struct pose_sample {
     vector6d_t p;
 };
 
+struct ephemeral_data_ {
```
acmorrow: nit: drop the trailing `_` since it is no longer private.

acmorrow (approved):
LGTM with one tiny naming nit. I may relocate / rename `ephemeral_data[_]` at some point down the line, but I'm happy enough with it just moving out for now.

PR #206, opened 2026-05-01
RSDK-13899 - Construct new kinematics on module startup from calibrated DH parameters for ur20 model

nfranczak:
There is a discrepancy between the TCP pose given to use through `GetEndPosition` and performing forward kinematics.
We close this discrepancy for the UR20 model.
On module startup we query the arm's controller for the the calibrated DH parameters and construct a brand new kinematics file (SVA type) on the fly which is then served through the `GetKinematics` endpoint.

TODO:

- [ ] once approved test in lab then in factory again to make sure everything works as expected.


src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -0,0 +1,521 @@
...
+    // Geometry poses are expressed in the *nominal* link's local frame -- the runtime
+    // builder applies a per-link correction so they land at the right world position
+    // when the chain is calibrated.
+    std::array<std::optional<Geometry>, 6> geometries;
+    NominalDH nominal;
+};
+
+// Per-model static tables. To register a new UR model, add an entry here.
+const std::map<std::string, ModelTables>& model_tables() {
+    static const std::map<std::string, ModelTables> tables = [] {
```
npmenard: not for this PR. Any way we can extract these from know kinematics files?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -493,6 +527,13 @@ viam::sdk::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
     const std::shared_lock rlock{config_mutex_};
     check_configured_(rlock);
 
+    if (!dh_kinematics_json_.empty()) {
```
npmenard: What i am concerned about is if the arm first sends the default kinematic file and then later the updated one then what happens? 
Since we read it during the config phase i think we  can't get into this scenario but wanted to check

src/viam/ur/module/ur_arm.cpp
```diff
@@ -403,6 +404,39 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
     VIAM_SDK_LOG(debug) << "URArm starting up";
     current_state_ = state_::create(configured_model_type, name(), cfg, ports_);
 
+    if (model_ == URArm::model("ur20")) {
+        VIAM_SDK_LOG(info) << "Fetching calibrated DH parameters from controller for ur20";
+        auto kin_info = current_state_->fetch_kinematics_info_();
```
npmenard: @Andrew is there a chance the configuration is attempted when a connection to the client was not established?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -663,6 +705,22 @@ ProtoStruct URArm::do_command(const ProtoStruct& command) {
                 cached_controlled_info->controlled = current_state_->is_current_state_controlled(&cached_controlled_info->description);
             }
             resp.emplace(k_get_state_description, cached_controlled_info->description);
+        } else if (kv.first == k_get_calibrated_dh_params) {
+            auto kin_info = current_state_->fetch_kinematics_info_();
```
npmenard: should cache the value, they are not expected to change right?

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -782,3 +782,25 @@ vector6d_t URArm::state_::clamp_limits_(vector6d_t desired, const std::optional<
...
+        [](auto& s) -> urcl::primary_interface::KinematicsInfo {
+            using S = std::decay_t<decltype(s)>;
+            if constexpr (std::is_base_of_v<state_connected_, S>) {
+                auto primary = s.arm_conn_->driver->getPrimaryClient();
+                if (!primary) {
+                    throw std::runtime_error("primary client unavailable on UrDriver");
+                }
+                auto kin = primary->getKinematicsInfo();
+                if (!kin) {
+                    throw std::runtime_error("controller has not delivered KinematicsInfo yet");
```
npmenard: can it fails transiently?

npmenard (approved):
LGTM i think this is good. just ask John/andrew a quick glance about the state machine changes if you haven't done so already
Left a couple of comments but nothing major

src/viam/ur/module/ur_arm.cpp
nfranczak: UR does not change its calibrated DH on the fly, so the kinematics that we build on module startup is valid.

src/viam/ur/module/ur_arm.cpp
nfranczak: will cache

src/viam/ur/module/ur_arm_state.cpp
nfranczak: should not but made it bullet proof.

CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ include(FetchContent)
 FetchContent_Declare(
   Universal_Robots_Client_Library
   GIT_REPOSITORY https://github.com/UniversalRobots/Universal_Robots_Client_Library
-  GIT_TAG        2.10.0
+  GIT_TAG        2.11.0
```
acmorrow, line 128: Neat: https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/496

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -8,6 +8,7 @@
 #include <thread>
 #include <variant>
 
+#include <ur_client_library/primary/robot_state/kinematics_info.h>
```
acmorrow: Please group this with the other ur_client_library includes below.

src/viam/ur/module/ur_arm.hpp
```diff
@@ -209,5 +210,16 @@ class URArm final : public Arm, public Reconfigurable {
...
+    // the DH-form JSON synthesized from the controller's DH parameters should
+    // be returned by get_kinematics().
+    std::string dh_kinematics_json_;
+
+    // Calibrated DH parameters captured once at configure_, reused by the
+    // get_calibrated_dh_params do_command. Calibration does not change at
+    // runtime, so caching the value avoids re-walking the state-machine
+    // variant and re-acquiring its mutex on every call.
+    std::optional<urcl::primary_interface::KinematicsInfo> calibrated_kin_info_;
+
```
acmorrow: These really should live on `state_`. That's where all the, well, state goes.

src/viam/ur/module/dh_kinematics.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <array>
+#include <string>
+
+namespace viam_ur {
```
acmorrow: I don't think we have a namespace in the rest of the project. I'd been meaning to add one, but it we do, I'd prefer to add one everywhere.  I'd say take this out for now.

src/viam/ur/module/dh_kinematics.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+// Per-joint DH parameters for a 6-DOF UR arm, as reported by the controller via
+// `urcl::primary_interface::KinematicsInfo`. Each array holds one entry per joint;
+// units match what urcl reports: meters for `a` and `d`, radians for `alpha` and
+// `theta`. `theta` is the per-joint calibration offset baked into the static link
+// pose -- it is NOT the joint variable (the joint angle is supplied by the
+// revolute joint frame in the synthesized chain).
+struct DHParams {
+    std::array<double, 6> a;
```
acmorrow: We have an alias for this: `vector6d`, which comes from URCL.

src/viam/ur/module/ur_arm.cpp
acmorrow: What if, say, the arm is shut down. We become disconnected. Calibration is changed offline. The arm is restarted and we reconnect?

I think the calibration info needs to live on `state_` and be re-initialized on reconnect.

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -0,0 +1,436 @@
...
+        ModelTables ur20;
+        ur20.limits = {{
+            {-360.0, 360.0},  // shoulder_pan
+            {-360.0, 360.0},  // shoulder_lift
+            {-180.0, 180.0},  // elbow
+            {-360.0, 360.0},  // wrist_1
+            {-360.0, 360.0},  // wrist_2
+            {-360.0, 360.0},  // wrist_3
+        }};
+        // Generated by tools/derive_dh_capsules.py -- do not edit by hand.
```
acmorrow: Where does this file live?

acmorrow (changes requested):
I think this needs some work. I think the flow of control between URArm and the state infrastructure to obtain the params should work differently: the calibrated dh results should live on the state_ object (or even in the `connected` state?). Callers wanting to obtain it should block on it becoming populated there. And the calibration I think needs to be re-fetched each time you reconnect: they could have changed during that offline period.

src/viam/ur/module/dh_kinematics.cpp
nfranczak: https://viam.atlassian.net/browse/RSDK-13902

src/viam/ur/module/dh_kinematics.cpp
nfranczak: old code when i first started this work. removing.

src/viam/ur/module/dh_kinematics.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <string>
+
+#include <ur_client_library/types.h>
+
+// Per-joint DH parameters for a 6-DOF UR arm, as reported by the controller via
+// `urcl::primary_interface::KinematicsInfo`. Each array holds one entry per joint;
+// units match what urcl reports: meters for `a` and `d`, radians for `alpha` and
+// `theta`. `theta` is the per-joint calibration offset baked into the static link
+// pose -- it is NOT the joint variable (the joint angle is supplied by the
+// revolute joint frame in the synthesized chain).
```
acmorrow: I don't really understand the last sentence of this comment. I've read it a few times and it is just confusing. I don't know what the "static link pose" means, and the whole negated thing is odd. What is a "joint variable"?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -403,6 +404,40 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
+    if (model_ == URArm::model("ur20")) {
+        VIAM_SDK_LOG(info) << "Fetching calibrated DH parameters from controller for ur20";
+        // Force the eager fetch+build at startup so any failure to obtain
+        // KinematicsInfo from the controller surfaces here rather than on the
+        // first get_kinematics() call. The state_ owns the cache and refreshes
+        // it on every reconnect.
+        const auto json = current_state_->get_dh_kinematics_json();
+        const auto kin_info = current_state_->get_calibrated_kinematics_info();
+
+        const auto fmt6 = [](const urcl::vector6d_t& v) {
```
acmorrow: Have a look at boost::io::make_ostream_joiner, used elsewhere in this file, and how it can be used with ranges, for a more compact way to write this.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -403,6 +404,40 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
+            }
+            os << "]";
+            return os.str();
+        };
+        VIAM_SDK_LOG(info) << "Calibrated KinematicsInfo received -- "
+                           << "a=" << fmt6(kin_info.dh_a_) << ", "
+                           << "d=" << fmt6(kin_info.dh_d_) << ", "
+                           << "alpha=" << fmt6(kin_info.dh_alpha_) << ", "
+                           << "theta=" << fmt6(kin_info.dh_theta_)
+                           << " (theta is baked into the static link pose; chain is emitted in SVA form)";
```
acmorrow: The parenthetical doesn't seem like it belongs in a log message

src/viam/ur/module/ur_arm.cpp
```diff
@@ -403,6 +404,40 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
+            for (size_t i = 0; i < v.size(); ++i) {
+                if (i) {
+                    os << ", ";
+                }
+                os << v[i];
+            }
+            os << "]";
+            return os.str();
+        };
+        VIAM_SDK_LOG(info) << "Calibrated KinematicsInfo received -- "
```
acmorrow: The `--` should probably be `:`.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -493,6 +528,19 @@ viam::sdk::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
...
     check_configured_(rlock);
 
+    auto dh_json = current_state_->get_dh_kinematics_json();
+    if (!dh_json.empty()) {
+        VIAM_SDK_LOG(info) << "get_kinematics: serving synthesized SVA JSON built from calibrated DH (" << dh_json.size() << " bytes)";
+        return viam::sdk::KinematicsDataSVA(std::vector<unsigned char>(dh_json.begin(), dh_json.end()));
+    }
+    // For models that are supposed to ship synthesized kinematics, refuse to fall through
+    // to the static file -- empty json here indicates a configuration bug.
+    if (model_ == model("ur20")) {
```
acmorrow, line 540: Why is this implemented only for UR20? Do the other UR arms work differently?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -663,6 +712,25 @@ ProtoStruct URArm::do_command(const ProtoStruct& command) {
                 cached_controlled_info->controlled = current_state_->is_current_state_controlled(&cached_controlled_info->description);
             }
             resp.emplace(k_get_state_description, cached_controlled_info->description);
+        } else if (kv.first == k_get_calibrated_dh_params) {
+            if (model_ != model("ur20")) {
+                throw std::runtime_error("calibrated DH parameters not available for this arm model");
+            }
+            const auto kin_info = current_state_->get_calibrated_kinematics_info();
+            auto to_array = [](const vector6d_t& v) {
```
acmorrow: See `add_limilts_response` above.

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -0,0 +1,432 @@
...
+    // One optional geometry per DH entry. nullopt means "no geometry on this segment".
+    // Geometry poses are expressed in the *nominal* link's local frame -- the runtime
+    // builder applies a per-link correction so they land at the right world position
+    // when the chain is calibrated.
+    std::array<std::optional<Geometry>, 6> geometries;
+    NominalDH nominal;
+};
+
+// Per-model static tables. To register a new UR model, add an entry here.
+const std::map<std::string, ModelTables>& model_tables() {
```
acmorrow: Unless you need ordered traversal, `std::unordered_map` is preferred.

src/viam/ur/module/dh_kinematics.cpp
acmorrow: OK, so, just to make sure I understand. We are deliberately hardcoding these for now just to demonstrate that this works, but we plan to extract these values from files for all models as the next step? I'm a little unclear on why we wouldn't just do this the right way up front.

src/viam/ur/module/dh_kinematics.hpp
acmorrow: I think in general some of the comments are very claude-y: it says something, but on closer inspection it is using a lot of terms without clear local meanings or antecedents. I don't think it is worth holding this review up over it, but in general, I think it is good practice to make sure that the comments it writes go through a round of refinement. Push back when it goes all `this thing is *term* -- it can't BE x`.

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -782,3 +783,131 @@ vector6d_t URArm::state_::clamp_limits_(vector6d_t desired, const std::optional<
...
+            [](auto& s) -> const arm_connection_* {
+                using S = std::decay_t<decltype(s)>;
+                if constexpr (std::is_base_of_v<state_connected_, S>) {
+                    return s.arm_conn_.get();
+                } else {
+                    return nullptr;
+                }
+            },
+            current_state_);
+        if (calibrated_kin_info_ && kin_cache_owner_ != nullptr && kin_cache_owner_ == current_conn) {
```
acmorrow: Unfortunately, I think this is undefined behavior. After the object pointed to by `kin_cache_owner_` is destroyed, even the pointer to it is now invalid and can't be used, even for a comparison.

Also, there is no guarantee that your newly allocated one can't land at the same address as the old one. In fact, it may well do so since you just handed the region back to the heap, but then went ahead and asked for another object of the same size!

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -782,3 +783,131 @@ vector6d_t URArm::state_::clamp_limits_(vector6d_t desired, const std::optional<
...
+                },
+                current_state_);
+        }
+        if (kin) {
+            return *kin;
+        }
+        if (std::chrono::steady_clock::now() >= deadline) {
+            throw std::runtime_error("controller has not delivered KinematicsInfo within timeout");
+        }
+        std::this_thread::sleep_for(k_poll_interval);
```
acmorrow: Let's talk about another way to architect this. I'm wondering if we can arrange things to work better with a future that gets fulfilled once the parameters arrive.

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -0,0 +1,432 @@
...
+        return t;
+    }();
+    return tables;
+}
+
+// Build a 4x4 homogeneous transform for a link's static pose:
+//   Rz(theta) * T(a, 0, d) * Rx(alpha)
+// All inputs in millimeters/radians. Output is the matrix that takes a child-frame
+// point to its parent-frame coordinates.
+Eigen::Matrix4d dh_link_pose_matrix(double a_mm, double d_mm, double alpha_rad, double theta_rad) {
```
acmorrow, line 167: To what extent do you want me to review the math in this file? It is possible that "none" is an acceptable answer, if you feel like you have done so (since I'm guessing this was mostly claude).

acmorrow (changes requested):
I still don't love the mechanics of how the cache works and I think it has some subtle undefined behavior.

I think some sort of notification based scheme would work better, probably via `future`. Let's talk about it in person maybe.

src/viam/ur/module/dh_kinematics.cpp
nfranczak: correct we are deliberately hardcoding these for now. In the follow up ticket I will remove this.

src/viam/ur/module/dh_kinematics.cpp
nfranczak: no need

src/viam/ur/module/ur_arm.cpp
nfranczak: the follow up ticket will handle the implementation for all other arm models.

src/viam/ur/module/ur_arm_state.cpp
acmorrow: @nfranczak - Please see https://github.com/acmorrow/universal-robots/commit/61c0e5a867780d93a57a7d47858851813b6bffaa for an alternative approach that I think works better. If you agree, I can push it to this PR or you can pull it in, either way.

nfranczak: I've pulled it in, thanks.

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -178,6 +217,29 @@ class URArm::state_ {
...
+    // the once_flag (rather than the whole payload) keeps the future's
+    // value type a plain `cached_kinematics_payload` and confines the heap
+    // indirection to a single field. A default member initializer for
+    // `json_once` lets the payload remain a C++20 aggregate so the producer
+    // can still construct it with designated initializers.
+    struct cached_kinematics_payload {
+        urcl::primary_interface::KinematicsInfo info;
+        std::string model_name;  // canonical lowercase, e.g. "ur20"
+        mutable std::unique_ptr<std::once_flag> json_once{std::make_unique<std::once_flag>()};
+        mutable std::string json{};
```
acmorrow: You are going to need to `// NOLINT(readability-redundant-member-init)` this, because right now you are caught between clang-tidy rejecting the "redundant" initializer, and GCC demanding it (and you need to make GCC happy by default).

acmorrow (approved):
LGTM!

PR #207, opened 2026-05-07
[RSDK-13667] Upgrade to trajex 0.100.0-rc5

acmorrow:
The counterpart to https://github.com/viam-modules/trajex/pull/11 which I will tag as rc5 once it merges.


acmorrow:
@npmenard - Hang back on this, the trajex release build failed, I am investigating. Unfortunately, I was overeager and tagged rc5, so this will need to be updated to a future rc6.

PR #209, opened 2026-05-19
RSDK-13902 - Augment model tables + fix 3dmodels endpoint

nfranczak:
Get calibrated DH parameters for all UR models.
Consume the existing  kinematics file and parse it into a model table which tells us the joint limits, geometries, and link names.
Reconstruct a new kinematics file using the calibrated DH parameters such that forward kinematics matches querying the controller for the TCP pose.


nfranczak:
For posterity:
Why did the 3dmodels endpoint break yesterday?

1. The names of links did not match up. glb names match with the link names in the ideal SVA while the SVA that we created on the fly used generic names like ur20_link0, ur20_link1, etc
2. There were seven glbs but only six links.
WRT 2: It is improper that there are seven glbs and a work around had to be introduced for this in #149

CMakeLists.txt
```diff
@@ -232,6 +232,7 @@ target_sources(viam-ur
     src/viam/ur/module/ur_arm_state_disconnected.cpp
     src/viam/ur/module/ur_arm_state_events.cpp
     src/viam/ur/module/ur_arm_state_independent.cpp
+    src/viam/ur/module/kinematics_parser.cpp
```
acmorrow: Should sort in higher up

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -842,8 +830,10 @@ std::string URArm::state_::get_dh_kinematics_json(std::chrono::steady_clock::dur
     // `json_once` is a `unique_ptr<once_flag>` (dereferenced here) because
     // `std::once_flag` is neither copyable nor movable.
     std::call_once(*payload.json_once, [&] {
+        const auto sva_path = resource_root_ / "kinematics" / (payload.model_name + ".json");
+        const ModelTables tbl = parse_kinematics(sva_path);
         const DHParams dh{payload.info.dh_a_, payload.info.dh_d_, payload.info.dh_alpha_, payload.info.dh_theta_};
-        payload.json = build_dh_kinematics_json(payload.model_name, dh);
+        payload.json = build_dh_kinematics_json(payload.model_name, dh, tbl);
```
acmorrow: You don't need to be so explicit with intermediates, btw. I'd expect the following to compile and mean the same thing:

```
payload.json = build_dh_kinematics_json(
    payload.model_name,
    {payload.info.dh_a_, payload.info.dh_d_, payload.info.dh_alpha_, payload.info.dh_theta_},
    parse_kinematics((resource_root_ / "kinematics" / payload.model_name).concat(".json")));
```

src/viam/ur/module/ur_arm_state_disconnected.cpp
```diff
@@ -88,13 +88,14 @@ std::unique_ptr<URArm::state_::arm_connection_> URArm::state_::state_disconnecte
...
-    // here.
+    // The dashboard reports the model in uppercase (e.g. "UR20") and in the
+    // coarser URCL category form ("ur5" covers both ur5e and ur7e), while
+    // `configured_model_type_` is the SDK model name in lowercase (e.g.
+    // "ur5e"). Lower-case the dashboard side and translate our SDK name to
+    // its URCL category for the comparison.
     boost::algorithm::to_lower(actual_model_type);
 
-    if (state.configured_model_type_ != actual_model_type) {
+    if (URArm::urcl_category(state.configured_model_type_) != actual_model_type) {
```
acmorrow: So, in `URArm::configure_` we convert the `model_` field (a C++ `Model` instance) into a std::string (by way of a very-much-no-longer-needed immediately-invoked-function-expression that dates from when `Model` lacked accessors). We then store that string in the `URArm::state_` instance. Then, here we feed that string back into `urcl_category`,  which appears to do a similar canonicalization? Or the inverse? I sort of can't tell.

And then there is a bunch of casing stuff going on, and special handling for `e` vs non-e and 7 vs 5.

It's getting more complex than I'm happy with, and I think it is setting us up for mistakes.

The right way to solve this is with a type that centralizes the logic. At minimum, we could start sending around the `Model` object rather than a stringified representation of it? Or we could make a wrapper around `Model` and adorn it with the relevant canonicalization and query operations that we need?

It has been bugging me for a while, but this may be the time to clean it up. The introduction of `urcl_category` I think is what pushes it over the edge.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -337,6 +337,21 @@ Model URArm::model(std::string model_name) {
     return {model_family(), std::move(model_name)};
 }
 
+std::string URArm::urcl_category(const std::string& sdk_model_name) {
```
acmorrow: Please see my thoughts below on this.

src/viam/ur/module/test.cpp
```diff
@@ -1097,3 +1099,192 @@ BOOST_AUTO_TEST_CASE(test_move_limit_vector_negative_element) {
...
 
 BOOST_AUTO_TEST_SUITE_END()
+
+BOOST_AUTO_TEST_SUITE(kinematics_parser_tests)
+
+namespace {
+
+std::filesystem::path test_kinematics_path(const std::string& model) {
+    // src/viam/ur/module/test.cpp -> src/kinematics/<model>.json
+    return std::filesystem::path{__FILE__}.parent_path().parent_path().parent_path().parent_path() / "kinematics" / (model + ".json");
```
acmorrow: I think this should be done more like how trajex does it with injecting something out of the build system. You can't assume the relative path between the cwd of the executing test and the source tree.

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,48 @@
...
+// joint limits for the six DH joints, the world-frame pose at zero joints of
+// each emitted link's geometry (capsule/sphere), and the seven emitted link
+// names in chain order.
+struct ModelTables {
+    std::array<JointLimits, 6> limits;
+    std::array<std::optional<Geometry>, 7> geometries;
+    std::array<std::string, 7> link_names;
+};
+
+// Parse a Viam-shipped SVA-form kinematics JSON describing a UR arm at zero
```
acmorrow: I'm not sure what `zero joints` means? Does that mean "all joint angles zero"? It says it above too

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,48 @@
...
+// Apply a 4x4 correction matrix to a viam-cpp-sdk pose, returning a new
+// pose. Internally round-trips through a quaternion (via rust-utils) to
+// compose with the matrix.
+viam::sdk::pose apply_correction_to_pose(const viam::sdk::pose& p, const Eigen::Matrix4d& correction);
+
+// Per-model data extracted from a shipped `kinematics/<model>.json` file:
+// joint limits for the six DH joints, the world-frame pose at zero joints of
+// each emitted link's geometry (capsule/sphere), and the seven emitted link
+// names in chain order.
+struct ModelTables {
```
acmorrow: Should ModelTables carry the model name (Model?) it describes with it?

src/viam/ur/module/dh_kinematics.hpp
```diff
@@ -16,18 +18,7 @@ struct DHParams {
...
-// the static link as a fixed Rz rotation -- DHParamConfig has no theta field,
-// and standard 4-parameter DH cannot represent UR's parallel-axis calibration
-// without resorting to numerically degenerate huge `d` values.
-//
-// Throws std::invalid_argument if model_name has no per-model table registered.
-std::string build_dh_kinematics_json(const std::string& model_name, const DHParams& dh);
+// Builds an RDK-compatible kinematics JSON document for the given UR model
+// from calibrated DH parameters plus a per-model table parsed from the
+// shipped `kinematics/<model>.json`.
+std::string build_dh_kinematics_json(const std::string& model_name, const DHParams& dh, const ModelTables& tbl);
```
acmorrow: - If `ModelTables` carried the model name, you wouldn't need it as a third param here.

- My undertanding is that `ModelTables` is the data derived statically from the files on disk, and `DHParams` is what we learned from talking to the arm. And we combine these to form the kinematics JSON. If that's true, I'd switch the order of the params: base truth first, corrections second.

- The above thought has me wondering something more profound though, about the design. Should the actual function be `ModelTables update_model_tables_with_dhparams(const ModelTables& tbl, const DHParams&)`? Or even give `ModelTables` a member function like `ModelTables ModelTables::apply_calibrations(const DHParms&)`? And then serialization is just about serializing the `ModelTables` data? Or is the result of applying `DHParams` to a `ModelTable` not representable as a `ModelTable`? Should it be?

src/viam/ur/module/kinematics_parser.cpp
```diff
@@ -0,0 +1,376 @@
...
+extern "C" void free_orientation_vector_memory(void* ov);
+extern "C" double* orientation_vector_get_components(void* ov);
+extern "C" void free_orientation_vector_components(double* ptr);
+extern "C" void* orientation_vector_from_quaternion(void* q);
+extern "C" void* quaternion_from_orientation_vector(void* ov);
+extern "C" void* quaternion_from_euler_angles(double roll, double pitch, double yaw);
+extern "C" void* new_quaternion(double real, double i, double j, double k);
+extern "C" void free_quaternion_memory(void* q);
+extern "C" double* quaternion_get_components(void* q);
+extern "C" void free_quaternion_components(double* ptr);
```
acmorrow: We also have these in `ur_arm.cpp`. I'd say write a `rust_utils.h` header for this, which both can use.

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+[[noreturn]] void throw_parse_error(const std::filesystem::path& path, const std::string& msg) {
+    throw std::invalid_argument("parse_kinematics: " + path.string() + ": " + msg);
+}
+
+// Returns the first numeric value among `keys` found in `obj`, or `fallback`
+// if none are present. Lets us absorb the x/y/z vs X/Y/Z divergence between
+// the shipped JSONs without committing to a normalization pass.
+double pick_double(const Json::Value& obj, std::initializer_list<const char*> keys, double fallback) {
+    for (const char* k : keys) {
```
acmorrow, line 83: You can use `std::string_view` in universal robots. On the other hand, that might trip up your initializer list thing here, which is working well. So maybe this is better.

src/viam/ur/module/kinematics_parser.cpp
```diff
@@ -0,0 +1,376 @@
...
+// Build a viam-cpp-sdk pose from an Eigen translation+quaternion pair by
+// converting the quaternion to an orientation-vector via rust-utils. The pose
+// stores its translation in mm (whatever unit the caller supplies) and its
+// `theta` field in degrees.
+viam::sdk::pose pose_from_translation_quaternion(const Eigen::Vector3d& t, const Eigen::Quaterniond& q) {
+    const std::unique_ptr<void, decltype(&free_quaternion_memory)> rust_q{new_quaternion(q.w(), q.x(), q.y(), q.z()),
+                                                                          &free_quaternion_memory};
+    const std::unique_ptr<void, decltype(&free_orientation_vector_memory)> rust_ov{orientation_vector_from_quaternion(rust_q.get()),
+                                                                                   &free_orientation_vector_memory};
+    const std::unique_ptr<double[], decltype(&free_orientation_vector_components)> comps{orientation_vector_get_components(rust_ov.get()),
```
acmorrow: You could probably make some helper functions for these.

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -0,0 +1,376 @@
...
+        throw_parse_error(sva_json_path, "missing `links` array");
+    }
+    if (!root.isMember("joints") || !root["joints"].isArray()) {
+        throw_parse_error(sva_json_path, "missing `joints` array");
+    }
+
+    // Build `parent_id -> child_entity` so we can walk the chain forward from
+    // "world" without ever looking up entries by name. Per the shipped JSONs
+    // each parent has at most one child (no branching); we enforce that here.
+    std::unordered_map<std::string, const Json::Value*> child_of;
```
acmorrow, line 443: Convince me there are no invalidation issues here, where the pointer in the map becomes pointing to something freed or moved.

acmorrow (changes requested):
Lots of thoughts on this one, lmk if you want to go over in person. Some are small, some are larger design questions.

My top two would be:

- Can we further simplify the model handling around a type and/or the C++ SDK `Model` class rather than doing string-y comparisons.

- Can we make this whole thing about round tripability to/from ModelTables, and then just treat DHParams as a way to mutate a ModelTables instance?

src/viam/ur/module/kinematics_parser.hpp
nfranczak: yes, I've updated the words to make it clearer.

src/viam/ur/module/dh_kinematics.hpp
nfranczak: I like the member function idea more.
The results of applying `DHParams` to a `ModelTable` is itself a `ModelTable`.

src/viam/ur/module/kinematics_parser.hpp
nfranczak: Good idea.

src/viam/ur/module/kinematics_parser.hpp
nfranczak: Fixed the ambiguous working.
Yes zero joints does mean all joint angles are are zero rotation.

src/viam/ur/module/ur_arm_state_disconnected.cpp
nfranczak: Introduced a `UrArmModel` class which would centralize the logic.

src/viam/ur/module/ur_arm_model.cpp
nfranczak: we turn the sva file into a container called root of type Json::Value
here root has a bunch of nodes where a node references a specific chunk of the original json we've read in, e.g. { "id": "base_link", "parent": "world" }
all nodes have some specific address like root["foo"][0]

child_of stores address of json nodes.
root exists while we construct child_of.

both root and child_of are local variables and they get destroyed in reverse order (https://stackoverflow.com/questions/14688285/c-local-variable-destruction-order)
so everytime we get an address from child_of there is something there to reference from in root.
This should handle the freed case.
The pointer will not point to something which no longer exists.

moving onto "moved".
For this to be an issue the following symptom would need to make itself apparent: a json node gets moved to a new address inside root after we've constructed it.
We use jsoncpp_lib (called jsoncpp) 
jsoncpp stores its nodes internally in std::map (https://github.com/open-source-parsers/jsoncpp/blob/master/include/json/value.h#L310) which had the following property: once a value is in it, its address does not change (https://stackoverflow.com/questions/61771352/will-it-change-the-address-of-a-existed-keys-value-when-inserting-new-keys)
Therefore we are immune to the "moved" symptom.

Hence invalidation issues should not occur.

CMakeLists.txt
```diff
@@ -289,6 +291,16 @@ target_link_libraries(universal-robots-test
...
+# Inject the absolute path to the shipped `kinematics/` data so the test
+# binary can load `<model>.json` regardless of the cwd it's invoked from
+# (otherwise tests have to derive paths from `__FILE__`, which breaks
+# under installed/relocated builds).
+target_compile_definitions(universal-robots-test
+  PRIVATE
+    VIAM_UR_TEST_KINEMATICS_DIR="${CMAKE_CURRENT_SOURCE_DIR}/src/kinematics"
+)
+
+apply_wall_werror(universal-robots-test)
```
acmorrow: This looks like maybe a bad merge?

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+struct ModelTable {
+    UrArmModel model;
+    std::array<JointLimits, 6> limits;
+    std::array<Eigen::Matrix4d, 7> link_locals;
+    std::array<std::optional<Geometry>, 7> geometries;
+    std::array<std::string, 7> link_names;
+
+    explicit ModelTable(UrArmModel m);
+
+    ModelTable with_calibrated_dh(const DHParams& dh) const;
```
acmorrow: Or `apply_calibration`?

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+};
+
+// Cumulative world-frame pose of the i-th link's parent at joints with zero rotation.
+Eigen::Matrix4d parent_pose_at(const ModelTable& tbl, std::size_t i);
```
acmorrow: This should be a const method on `ModelTable`

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+// Apply a 4x4 correction matrix to a viam-cpp-sdk pose, returning a new
+// pose. Internally round-trips through a quaternion (via rust-utils) to
+// compose with the matrix.
+viam::sdk::pose apply_correction_to_pose(const viam::sdk::pose& p, const Eigen::Matrix4d& correction);
+
+// Parse a Viam-shipped SVA-form kinematics JSON describing a UR arm at
+// joints with zero rotation.
+ModelTable parse_kinematics(const std::filesystem::path& sva_json_path, UrArmModel arm_model);
```
acmorrow: This could be a static factory method on `ModelTable`, something like `static ModelTable ModelTable::from_sva_json(const std::filesystem::path& sva_json_path, UrArmModel arm_model)`.

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+// Serialize a ModelTable to an RDK-compatible SVA kinematics JSON.
+std::string to_sva_json(const ModelTable& tbl);
```
acmorrow: This could be a non-static const method on `ModelTable`, like `std::string ModelTable::to_sva_json()`;

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -181,252 +32,46 @@ Eigen::Matrix4d dh_link_pose_matrix(double a_mm, double d_mm, double alpha_rad,
...
-                r.tz_mm = out.tz;
-                return r;
-            }
-        },
-        geom);
+Eigen::Matrix4d dh_local_for_slot(std::size_t i, const DHParams& dh) {
+    if (i == 0) {
+        return dh_link_pose_matrix(0.0, dh.d[0] * k_m_to_mm, 0.0, 0.0);
+    }
+    const std::size_t s = i - 1;
```
acmorrow: I don't understand the special casing around 0 here. And won't it be accessing the same `dh.d` (at 0) for i=0 and i=1? At minimum I think a comment is warranted.

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -181,252 +32,46 @@ Eigen::Matrix4d dh_link_pose_matrix(double a_mm, double d_mm, double alpha_rad,
...
+    const double d_mm = (i == 1) ? 0.0 : (dh.d[s] * k_m_to_mm);
```
acmorrow: You've just created `s` which is a shifted `i`, so it is weird to see `i` reappear here. This would be clearer as `(s == 0)` I think

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -181,252 +32,46 @@ Eigen::Matrix4d dh_link_pose_matrix(double a_mm, double d_mm, double alpha_rad,
...
-                const PoseTQ out = apply_correction(correction, in);
-                SphereGeometry r = g;
-                r.tx_mm = out.tx;
-                r.ty_mm = out.ty;
-                r.tz_mm = out.tz;
-                return r;
-            }
-        },
-        geom);
+Eigen::Matrix4d dh_local_for_slot(std::size_t i, const DHParams& dh) {
```
acmorrow: Better, usually, to write something like this with params ordered `(const DHParams& dh, std::size_t i)`, since that's in the order "thing I'm querying, thing I'm querying for in it".

src/viam/ur/module/dh_kinematics.cpp
```diff
@@ -181,252 +32,46 @@ Eigen::Matrix4d dh_link_pose_matrix(double a_mm, double d_mm, double alpha_rad,
...
-    const auto it = tables.find(model_name);
-    if (it == tables.end()) {
-        throw std::invalid_argument("build_dh_kinematics_json: no per-model table for `" + model_name + "`");
-    }
-    const ModelTables& tbl = it->second;
-
-    Json::Value root(Json::objectValue);
-    root["name"] = model_name;
-    root["kinematic_param_type"] = "SVA";
+ModelTable ModelTable::with_calibrated_dh(const DHParams& dh) const {
```
acmorrow: If these are declared in `kinematics_parser.hpp`, then they should be defined in `kinematics_parser.cpp`.

src/viam/ur/module/dh_kinematics.hpp
```diff
@@ -1,33 +1,15 @@
...
-// Per-joint DH parameters for a 6-DOF UR arm, as reported by the controller via
-// `urcl::primary_interface::KinematicsInfo`. Each array holds one entry per joint;
-// units match what urcl reports: meters for `a` and `d`, radians for `alpha` and
-// `theta`. The reported `theta` is a fixed per-joint calibration offset; it does
-// not change as the joint moves.
+// Per-joint DH parameters for a 6-DOF UR arm, as reported by the controller
+// via `urcl::primary_interface::KinematicsInfo`. Each array holds one entry
+// per joint; units match what urcl reports: meters for `a` and `d`, radians
+// for `alpha` and `theta`. The reported `theta` is a fixed per-joint
+// calibration offset; it does not change as the joint moves.
```
acmorrow: If this is all that is left of the `DHParams` header, maybe it just belongs in `kinematics_parser.hpp`?

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+// A geometry expressed in its link's parent (joint) frame.
+// `pose` translation in mm, orientation as an OV unit-vector with
+// `theta` in degrees. `shape` is one of viam-cpp-sdk's shape primitives.
+struct Geometry {
+    viam::sdk::pose pose;
+    std::variant<viam::sdk::sphere, viam::sdk::capsule> shape;
+};
+
+// Mirror of an RDK SVA kinematics document.
+struct ModelTable {
```
acmorrow: Or maybe `ModelKinematics`? Or even `UrArmModel::kinematics`?

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <variant>
+
+#include <Eigen/Geometry>
+
+#include <viam/sdk/common/pose.hpp>
+#include <viam/sdk/spatialmath/geometry.hpp>
+
+#include "ur_models.hpp"
+
+struct DHParams;
```
acmorrow: Should this just be a struct nested in `ModelKinematics`?

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
...
+    ModelTable with_calibrated_dh(const DHParams& dh) const;
+};
+
+// Cumulative world-frame pose of the i-th link's parent at joints with zero rotation.
+Eigen::Matrix4d parent_pose_at(const ModelTable& tbl, std::size_t i);
+
+// Apply a 4x4 correction matrix to a viam-cpp-sdk pose, returning a new
+// pose. Internally round-trips through a quaternion (via rust-utils) to
+// compose with the matrix.
+viam::sdk::pose apply_correction_to_pose(const viam::sdk::pose& p, const Eigen::Matrix4d& correction);
```
acmorrow: Is this actually user facing? Or should it be an implementation detail.

src/viam/ur/module/ur_models.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+struct UrModelDescriptor {
+    std::string sdk_name;
+    std::string urcl_category;
+    std::vector<std::string> mesh_parts;
+};
+
+// Returns the descriptor for an SDK model name. Throws
+// std::invalid_argument if the name is not registered.
+const UrModelDescriptor& ur_model_descriptor(const std::string& sdk_name);
```
acmorrow: Could just be a static function on `UrModelDescriptor`, like `static const UrModelDescriptor& for_sdk_name(const std::string& sdk_name);

src/viam/ur/module/ur_models.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+// Returns all registered descriptors, in registration order. Used by
+// `URArm::create_model_registrations` so adding a new arm is a single-site
+// change here.
+const std::vector<UrModelDescriptor>& ur_model_descriptors();
```
acmorrow: Could also be a static function on `UrModelDescriptor`, like `const std::vector<UrModelDescriptor>& get_registered()` or similar.

src/viam/ur/module/ur_arm_model.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+    // Build from an SDK model name (e.g. "ur5e"). Synthesizes the SDK
+    // Model in `URArm::model_family()`. Throws std::invalid_argument if
+    // the name is not registered.
+    static UrArmModel from_sdk_name(const std::string& sdk_name);
+
+    const viam::sdk::Model& sdk_model() const& {
+        return sdk_model_;
+    }
+    const std::string& sdk_name() const& {
```
acmorrow, line 85: Please throw a blank line after the `}` before declaring the next method.

src/viam/ur/module/ur_models.hpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+    const std::string& sdk_name() const& {
+        return descriptor_->sdk_name;
+    }
+    const std::string& urcl_category() const& {
+        return descriptor_->urcl_category;
+    }
+    const std::vector<std::string>& mesh_parts() const& {
+        return descriptor_->mesh_parts;
+    }
```
acmorrow: I'm not sure these two functions are necessary. Once you have the result of calling `descriptor`, these are just field access.

src/viam/ur/module/kinematics_parser.hpp
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
acmorrow: The contents of this file isn't really accurately captured by calling it "kinematics parser", or, at least not anymore.

src/viam/ur/module/ur_models.cpp
```diff
@@ -0,0 +1,56 @@
...
+    for (const auto& d : descriptors()) {
+        if (d.sdk_name == sdk_name) {
+            return d;
+        }
+    }
+    throw std::invalid_argument("ur_model_descriptor: unknown sdk model name `" + sdk_name + "`");
+}
+
+const std::vector<UrModelDescriptor>& ur_model_descriptors() {
+    return descriptors();
```
acmorrow: Why do this in a separate function rather than just declaring the `static const table` here?

acmorrow (changes requested):
This is coming along nicely, but I still see many places where types are being created, and then functions that act on them, rather than making those methods of the type, either as statics for factories, or instance members for the rest.

Also, now that UrArmModel exists, it seems to me like it is the clear home for the kinematics data and associated methods.

Let me know if you want to go over any of it in person.

src/viam/ur/module/dh_kinematics.cpp
nfranczak: By the time that we enter `dh_local_for_slot` all arm models are converted into a model table composed of 7 entries. 
I'ts 7 on purpose because there are 6 articulated links plus 1 base link (for ur5e and ur20 there are also 7 glbs).

However there are 6 DH rows since there are only 6 joints.

And so, the base link needs some local transform.
So, we treat d[0] (the vertical translation of joint 1 (shoulder joint)) as the base / shoulder link offset (there is no base joint, the first joint that appears is the shoulder joint). (Note: the chronology of 

So d[0] is used in the 0th entry of the model table. Then the next entry in the model table still uses all the other DH param values at index 0 like alpha, theta, and a, but d is zero'd out so that it's not applied twice.

src/viam/ur/module/kinematics_parser.hpp
nfranczak: removed from the hpp file entirely.

src/viam/ur/module/kinematics_parser.hpp
nfranczak: I prefer the name `apply_calibrated_dh`

src/viam/ur/module/kinematics_parser.hpp
nfranczak: Lets do ModelKinematics.

src/viam/ur/module/kinematics_parser.hpp
nfranczak: hmm, this could go either way.

The arm's controller is what reports DHParams and ModelKinematics is its consumer.
So it being nested would give the implication that DHparams belong to the definition of ModelKineamtics when in fact they are just used by one of the functions (apply_calibrated_dh)

I lean on the side of keeping as is.

src/viam/ur/module/kinematics_parser.hpp
nfranczak: Will rename to `model_kinematics.hpp`

src/viam/ur/module/ur_arm_state_disconnected.cpp
```diff
@@ -88,15 +88,15 @@ std::unique_ptr<URArm::state_::arm_connection_> URArm::state_::state_disconnecte
     if (!arm_connection->dashboard->commandGetRobotModel(actual_model_type)) {
         throw std::runtime_error("failed to get model info of connected arm");
     }
-    // The dashboard reports the model in uppercase (e.g. "UR20") while
-    // `configured_model_type_` is normalized to lowercase at the assignment
-    // site in `URArm::configure_`. Compare case-insensitively by lowering
-    // here.
+    // The dashboard reports the model in uppercase (e.g. "UR20") in the
+    // coarser URCL category form ("ur5" covers both ur5e and ur7e), while
```
acmorrow: Shouldn't the `"ur5"` in the parenthetical on line 92 be capitalized: `"UR5"`?

src/viam/ur/module/ur_models.hpp
```diff
@@ -0,0 +1,137 @@
...
+// it was parsed for. Constructed via `UrArmModel::load_kinematics(path)`.
+class UrArmModel::Kinematics {
+   public:
+    UrArmModel model;
+    std::array<JointLimits, 6> limits;
+    std::array<Eigen::Matrix4d, 7> link_locals;
+    std::array<std::optional<Geometry>, 7> geometries;
+    std::array<std::string, 7> link_names;
+
+    Kinematics apply_calibrated_dh(const DHParams& dh) const;
```
acmorrow: Or just `calibrate_with`? The `DHParams` aren't the things that are calibrated (e.g. the thing we are applying). They ARE the calibration. We are calibrating ourselves with the given DHParams.

src/viam/ur/module/ur_arm_model.hpp
```diff
@@ -0,0 +1,137 @@
+#pragma once
```
acmorrow, line 1: Filename should probably match the class, so `ur_arm_model.hpp` / `ur_arm_model.cpp`, but save that for right before you merge since otherwise it just makes reviewing harder.

CMakeLists.txt
```diff
@@ -231,6 +231,7 @@ target_sources(viam-ur
     src/viam/ur/module/ur_arm_state_disconnected.cpp
     src/viam/ur/module/ur_arm_state_events.cpp
     src/viam/ur/module/ur_arm_state_independent.cpp
+    src/viam/ur/module/ur_models.cpp
```
acmorrow: - `ur_models.cpp` to `ur_arm_model.cpp` per above.
- `ur_model_kinematics.cpp` goes away, folded into `ur_arm_model.cpp`
-  Remember to sort the file in, `ur_arm_model.cpp` would land between `ur_arm.cpp` and `ur_arm_state.cpp`.

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -842,8 +829,9 @@ std::string URArm::state_::get_dh_kinematics_json(std::chrono::steady_clock::dur
     // `json_once` is a `unique_ptr<once_flag>` (dereferenced here) because
     // `std::once_flag` is neither copyable nor movable.
     std::call_once(*payload.json_once, [&] {
-        const DHParams dh{payload.info.dh_a_, payload.info.dh_d_, payload.info.dh_alpha_, payload.info.dh_theta_};
-        payload.json = build_dh_kinematics_json(payload.model_name, dh);
+        payload.json = payload.arm_model.load_kinematics(resource_root_ / "kinematics" / (payload.arm_model.sdk_name() + ".json"))
```
acmorrow: If you do `.concat(".json")` instead of using string concatenation, you stay in filesystem API the whole time: `...load_kinematics((resource_root_ / "kinematics" / (payload.arm_model.sdk_name()).concat(".json")`

src/viam/ur/module/ur_models.cpp
```diff
@@ -0,0 +1,48 @@
+#include "ur_models.hpp"
+
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "ur_arm.hpp"  // URArm::model_family()
```
acmorrow: Personally, I don't like having the "why was this #include'd" comments. What if you want more than one thing from it? Or what you are getting from it changes names? They invariably are incomplete, or stale. Better to just let the include speak for itself: I need stuff from here.

src/viam/ur/module/ur_models_kinematics.cpp
```diff
@@ -0,0 +1,539 @@
+#include "ur_models.hpp"
```
acmorrow: The contents of this file belong in `ur_arm_model.cpp` (currently called `ur_models.cpp`)

src/viam/ur/module/ur_arm_model.cpp
acmorrow: But, if jsoncpp should ever decide to change its internal representation, that property might cease to hold, right?

src/viam/ur/module/ur_arm.cpp
```diff
@@ -356,16 +322,12 @@ std::vector<std::shared_ptr<ModelRegistration>> URArm::create_model_registration
             [](auto const& config) { return validate_config_(config); });
     };
 
-    auto registrations = model_strings | boost::adaptors::transformed(registration_factory);
+    auto registrations = UrModelDescriptor::all() | boost::adaptors::transformed(registration_factory);
```
acmorrow, line 325: I love it

src/viam/ur/module/ur_arm.cpp
```diff
@@ -377,29 +339,6 @@ void URArm::configure_(const std::unique_lock<std::shared_mutex>& lock, const De
...
-    // check model type is valid, map to ur_client data type
-    // https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/bff7bf2e2a85c17fa3f88adda241763040596ff1/include/ur_client_library/ur/datatypes.h#L204
-    //
-    // The values returned here are stored verbatim into
-    // `state_::configured_model_type_` and compared elsewhere (against
-    // `actual_model_type` reported by the dashboard, and against literal
-    // model-name strings in the kinematics readers and `model_tables()`
-    // lookups). The canonical project-wide spelling is lowercase, so we
-    // produce lowercase here at the assignment site.
-    const std::string configured_model_type = [&] {
```
acmorrow, line 389: So happy to see this go away.

src/viam/ur/module/ur_arm.cpp
```diff
@@ -527,53 +464,13 @@ viam::sdk::KinematicsData URArm::get_kinematics(const ProtoStruct&) {
...
-    // Read the entire file into a vector without computing size ahead of time
-    std::vector<char> temp_bytes(std::istreambuf_iterator<char>(sva_file), {});
-    if (sva_file.bad()) {
-        throw std::runtime_error(str(boost::format("error reading kinematics file '%1%'") % sva_file_path));
-    }
-
-    // Convert to unsigned char vector
-    return viam::sdk::KinematicsDataSVA({temp_bytes.begin(), temp_bytes.end()});
+    VIAM_SDK_LOG(info) << "get_kinematics: serving synthesized SVA JSON built from calibrated DH (" << dh_json.size() << " bytes)";
+    return viam::sdk::KinematicsDataSVA(std::vector<unsigned char>(dh_json.begin(), dh_json.end()));
```
acmorrow, line 473: Love this too

acmorrow (changes requested):
This is looking great. Mostly only small / mechanical things. I think one last round and you will have it. You can hold off on any file moves / renames until after we approve in this layout. Just don't forget to do them before merge.

src/viam/ur/module/ur_arm_state_disconnected.cpp
nfranczak: yes

src/viam/ur/module/ur_models.hpp
nfranczak: ok

src/viam/ur/module/ur_models_kinematics.cpp
nfranczak: ok, won't do it now though per one of your other comments.

src/viam/ur/module/ur_arm_model.cpp
acmorrow: Let me ask a different question, to maybe resolve this cleanly: does the jsoncpp data structure that `child_of` holds pointers into ever get mutated while `child_of` is being used? Or is it strictly read-only during what amounts to a validation pass? If the latter, then I'm satisfied (though there should likely be a comment explaining that restriction on the use of `child_of`.

acmorrow (approved):
This LGTM, with one open question about the `child_of` data structure that I can't quite tell the answer to while looking at the code. Depending on that answer I think you can go ahead and merge, but don't forget to do the file renames / merges I mentioned.

src/viam/ur/module/ur_arm_model.cpp
nfranczak: > does the jsoncpp data structure that child_of holds pointers into ever get mutated while child_of is being used? Or is it strictly read-only during what amounts to a validation pass?

Yes, the jsoncpp data structure that `child_of` holds pointers into (named `root`) is **strictly read-only**. There is no augmentation of that data structure.

acmorrow: I don't know the jsoncpp library well enough to know if it propagates `const` correctly so that mutation of a JSON document is prohibited via a `const JsonValue`, but if so, you can use an immediately evaluated function expression to bring `root` into being as a `const` object despite needing to "mutate" it with `parseFromStream`:

```
const auto root = [&]() {
    Json::Value root;
    const Json::CharReaderBuilder reader_builder;
    std::string errs;
    if (!Json::parseFromStream(reader_builder, in, &root, &errs)) {
        throw_parse_error(sva_json_path, "JSON parse failure: " + errs);
    }
    return root;
}();
```

nfranczak:
Note: Need to test this out on hardware (which I will do today) before merging.

src/viam/ur/module/ur_arm_model.cpp
nfranczak: hmm, I asked claude to do some investigation on this. This was it's response. Seems that **_yes it propagates `const` correctly._**
```
The proof is in jsoncpp's own header (/usr/include/jsoncpp/json/value.h). It declares two overloads of operator[], with the documentation explicitly distinguishing mutation behavior:

  // value.h:472-478
  /// Access an object value by name, create a null member if it does not exist.
  Value& operator[](const char* key);                     // non-const → can mutate

  /// Access an object value by name, returns null if there is no member with that name.
  const Value& operator[](const char* key) const;         // const → read-only

  And similarly for array indexing at lines 440-454:

  /// Access an array element (zero based index). If the array contains less
  /// than index element, then null value are inserted in the array so that
  /// its size is index+1.
  Value& operator[](ArrayIndex index);                    // non-const → extends with nulls

  /// Access an array element (zero based index).
  const Value& operator[](ArrayIndex index) const;        // const → read-only

  The documentation is explicit: the non-const overloads insert null members / extend arrays, while the const overloads do neither.

  When root is declared const auto root = ..., every root[...] call site resolves via overload resolution to the const version, which returns const Value& and cannot mutate. The compiler enforces this — any root["new_key"] = something would fail to
  compile because there's no non-const overload available on a const receiver.

  So the const-ness of root end-to-end guarantees the no-mutation invariant that child_of's pointer validity depends on. That's what makes the IIFE pattern stronger than the comment I had before — the comment was a request to humans; this is a
  contract with the compiler.

```

PR #210, opened 2026-05-22
[RSDK-13667] Upgrade trajex to v0.100.0

acmorrow:
@npmenard - I anticipate tagging this commit as Universal Robots 0.6.0 GA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ endif()
 option(VIAM_UR_USE_WALL_WERROR "Build with -Wall and -Werror flags" ON)
 
 # Function to apply strict warning flags to a target
-function(apply_wall_werror target_name)
+function(viam_ur_apply_wall_werror target_name)
   if (NOT VIAM_UR_USE_WALL_WERROR)
     return()
   endif()
@@ -249,7 +249,7 @@ target_link_libraries(viam-ur
     jsoncpp_lib
 )
 
-apply_wall_werror(viam-ur)
+viam_ur_apply_wall_werror(viam-ur)
 
 install(
   TARGETS
@@ -269,7 +269,7 @@ target_link_libraries(universal-robots
     viam-ur
 )
 
-apply_wall_werror(universal-robots)
+viam_ur_apply_wall_werror(universal-robots)
 
 install(
   TARGETS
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -141,7 +141,7 @@ FetchContent_Declare(
   # in failed compilation due to the statically included protobuf gens
   # getting in the wrong order. Just delete them, we don't need them.
   PATCH_COMMAND find ./src/viam/api -name "*.pb.h" -type f -exec rm {} +
-  FIND_PACKAGE_ARGS
+  FIND_PACKAGE_ARGS 0.31.0 CONFIG COMPONENTS viamsdk
```
acmorrow, line 144: The `find_package(viam-cpp-sdk` below was kinda wrong. The right way to combine FetchContent with `find_package` is as we have it now, where the arguments you want FetchContent to use when it looks for a satisfying system package is to place them here. Once you have done that, a secondary call to `find_package` is not needed, so I've removed it below.

CMakeLists.txt
```diff
@@ -167,9 +165,10 @@ find_package(Boost CONFIG REQUIRED
 FetchContent_Declare(
   trajex
   GIT_REPOSITORY https://github.com/viam-modules/trajex.git
-  GIT_TAG v0.100.0-rc6
+  GIT_TAG v0.100.0
   GIT_SHALLOW TRUE
 )
+set(VIAM_TRAJEX_ENABLE_TOOLS ON)
```
acmorrow, line 171: After https://github.com/viam-modules/trajex/pull/13, trajex does not enable the tools (aka planner) when embedded via FetchContent, so we need to ask for them.

When legacy support drops out of trajex, this goes away.

CMakeLists.txt
```diff
@@ -167,9 +165,10 @@ find_package(Boost CONFIG REQUIRED
 FetchContent_Declare(
   trajex
   GIT_REPOSITORY https://github.com/viam-modules/trajex.git
-  GIT_TAG v0.100.0-rc6
+  GIT_TAG v0.100.0
```
acmorrow, line 168: 🎉

npmenard (approved):

PR #211, opened 2026-05-28
RSDK-14032 - update viam-cpp-sdk to v38

nfranczak:


CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ FetchContent_MakeAvailable(Universal_Robots_Client_Library)
 FetchContent_Declare(
   viam-cpp-sdk
   GIT_REPOSITORY https://github.com/viamrobotics/viam-cpp-sdk
-  GIT_TAG releases/v0.31.0 # Keep in sync with Dockerfile
+  GIT_TAG releases/v0.37.0 # Keep in sync with Dockerfile
```
acmorrow: Below, you need it on `FIND_PACKAGE_ARGS` too.

acmorrow (changes requested):

acmorrow (approved):
LGTM, however, please do not merge this yet. We should land #209 first so we can tag it as 0.6.1. If we do it the other way around, 0.6.1 won't build because the docker image will be too new for it.

acmorrow:
@nfranczak -

We are going to address the pinning issue by way of #212. We will land that first, then #209. That will set us up for an 0.6.1-rc1 that is reproducible.

Then, we will handle #211. That will need to get split into two commits: an update to the docker file, which will trigger a docker rebuild, and then the main commit which upgrades TO the new docker file. I'lll leave more comments on #211

acmorrow:
> I'lll leave more comments on #211

Lol this is #211. Moving too fast ...

Anyway, here is what should be done. Once we have #212 and #209 landed, then this PR should be split.

1) Pull out ONLY the Dockerfile changes into a new PR. I'll instantly approve it, and you can merge.
2) Wait for CI to build and push the new Docker images.
3) Remove the Dockerfile changes from this PR. CI should fail, because the pinned Docker image is out of date.
4) Update the SHAs in this PR to use the new images.
5) Confirm CI now passes.

Then we can merge this safely, and everything after #212 will now be reproducible in CI.

Dockerfile
```diff
@@ -91,7 +91,7 @@ RUN cd /root/opt/src && \
 RUN cd /root/opt/src && \
     git clone https://github.com/viamrobotics/viam-cpp-sdk && \
     cd viam-cpp-sdk && \
-    git checkout releases/v0.31.0 && \
```
acmorrow: Please split this out into a new, standalone PR, rebased above the commit from #212.

acmorrow (changes requested):

acmorrow (approved):
LGTM if CI passes

PR #219, opened 2026-06-17
RSDK-14141 Add zero_ftsensor DoCommand to zero the wrist force/torque sensor

vijayvuyyuru:
## Summary

Adds a no-argument `zero_ftsensor` arm `DoCommand` that tares the UR e-Series wrist force/torque sensor at the controller. After it runs, the existing `get_tcp_forces_base` / `get_tcp_forces_tool` readings report ~0 in the current pose, giving consumers a one-call hardware tare over the same `DoCommand` surface instead of doing a software offset-subtraction or taring out-of-band from the teach pendant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,16 @@ example output:
 }
 ```
 
+#### zero_ftsensor DoCommand
+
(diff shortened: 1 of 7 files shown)


acmorrow (approved):
LGTM. When you squash merge, please put the ticket number in brackets in the commit, like `[RSDK-14141] Add zero_ftsensor ... `.

PR #220, opened 2026-06-25
Add UR12e support

JosephBorodach:
Adds `viam:universal-robots:ur12e`, mapping to URCL's `UR10` category. Kinematics from UR's official ROS2 description (`config/ur12e/default_kinematics.yaml`).

## Manual Testing 
Validated against a real UR12e


JosephBorodach:
@acmorrow @nfranczak

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -338,6 +338,9 @@ const std::vector<UrModelDescriptor>& UrModelDescriptor::all() {
         t.push_back({"ur20",
                      "ur20",
                      {"base_link", "wrist_3_link", "shoulder_link", "forearm_link", "upper_arm_link", "wrist_1_link", "wrist_2_link"}});
+        // URCL maps "ur12e" to RobotType::UR10 (per URCL helpers.h), so the
+        // urcl_category here is "ur10".
+        t.push_back({"ur12e", "ur10", {}});
```
nfranczak, line 347: third list cannot be empty.

nfranczak (commented):
please perform the following check
1. get the kinematics from the robot through a client script
2. perform forward kinematics (if you're using golang this is model.transform on some inputs)
3. get the current end effector postion through the arm api 
4. calculate the delta
delta should be on the order of 1-500 microns for position

acmorrow:
@JosephBorodach - Could you please file a ticket for this, so that we can track what release support lands in?

acmorrow (commented):
This LGTM when @nfranczak is happy with it. Please do file a ticket, and include the ticket on the commit message like `[RSDK-ABCDE] Add UR12e support`.

src/viam/ur/module/ur_arm_model.cpp
nfranczak: thought about it, I think that it can be empty since the get3d models endpoint is not implemented for this arm model as it  is new.

please file a ticket as a todo and add it as a comment here that the third list will need to be updated the day that the relevant glbs are added.

nfranczak:
lets also add a urdf file for the ur12e

JosephBorodach:
> Joseph Borodach (@JosephBorodach) - Could you please file a ticket for this, so that we can track what release support lands in?

https://viam.atlassian.net/browse/RSDK-14278?atlOrigin=eyJpIjoiNWZmZDY0Mjk0MGVlNGFjZGIzMDNjZmFlNzRkZjRiOGUiLCJwIjoiaiJ9

src/viam/ur/module/ur_arm_model.cpp
JosephBorodach: https://viam.atlassian.net/browse/RSDK-14279?atlOrigin=eyJpIjoiN2YxOGU4MDg4ZjUzNDkzMjg0NjRjOTg2MmY5YmRkODMiLCJwIjoiaiJ9

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -338,6 +338,13 @@ const std::vector<UrModelDescriptor>& UrModelDescriptor::all() {
         t.push_back({"ur20",
                      "ur20",
                      {"base_link", "wrist_3_link", "shoulder_link", "forearm_link", "upper_arm_link", "wrist_1_link", "wrist_2_link"}});
+        // URCL maps "ur12e" to RobotType::UR10 (per URCL helpers.h), so the
+        // urcl_category here is "ur10".
+        // TODO(RSDK-XXXXX): populate the mesh-link list here once ur12e GLBs
```
acmorrow, line 343: This can now say RSDK-14279

meta.json
```diff
@@ -0,0 +1,40 @@
+{
```
acmorrow, line 1: This file should be getting generated, not checked in.

src/kinematics/ur12e.json
```diff
@@ -0,0 +1,250 @@
+{
```
acmorrow, line 1: What's the provenance of this file / the values it contains?

src/kinematics/ur12e.urdf
```diff
@@ -0,0 +1,222 @@
+<?xml version="1.0" ?>
```
acmorrow, line 1: What's the provenance of this file / the values it contains?

src/viam/ur/module/ur_arm_model.cpp
acmorrow: @JohnN193 - Curious for your take on this. If UR10 is the "real" model, should all of this be getting placed in under a UR10 identifier then we just map ur12e to that? I can't remember exactly what the desired arrangement would be.

acmorrow (changes requested):

src/viam/ur/module/ur_arm_model.cpp
JohnN193: the current setup matches what we did for the ur5e & ur7e. technically the geometry might be different, but i haven't looked into how different it will be

JohnN193: looking closer at the UR's [ros descriptions](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/rolling/config/ur12e/joint_limits.yaml#L13C1-L13C35) it does seem like they just reuse the ur5e/[ur10e](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/89bbe795f38a7ab00fb66fe8831dfff79dc99edf/config/ur12e/visual_parameters.yaml#L12), so maybe we should just treat this all as a ur10

PR #221, opened 2026-07-02
ci: add workflow to auto-bump viam-cpp-sdk

stuqdog:
Adds `.github/workflows/bump-viam-cpp-sdk.yml` — a daily scheduled workflow (plus `workflow_dispatch`) that opens/updates a PR bumping `viam-cpp-sdk` whenever a newer release is published on [viamrobotics/viam-cpp-sdk](https://github.com/viamrobotics/viam-cpp-sdk/releases).

Why a custom workflow rather than Dependabot: the SDK is pinned via **conan** (no Dependabot ecosystem) **and** the version is duplicated as a git tag across several build files, so it needs multi-file syncing.

The workflow updates **every** location in one PR and **fails loudly** if any old reference is left behind (rather than opening a half-updated PR):
- `releases/vX.Y.Z` git tags in `Dockerfile`, `bin/setup.sh`, `CMakeLists.txt` (auto-discovered via `git grep`)
- the `viam-cpp-sdk/[>=X.Y.Z]` conan range in `conanfile.py`
- the CMake find version (`FIND_PACKAGE_ARGS X.Y.Z` / `find_package(viam-cpp-sdk X.Y ...)`)

Uses `perl -i` (consistent across runners) and only bumps forward. Opens the PR with the org `GIT_ACCESS_TOKEN` bot secret (same one `viam-python-sdk`'s `update_protos` uses) so the bump PR triggers CI.

Opened as a draft for review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/bump-viam-cpp-sdk.yml
+++ b/.github/workflows/bump-viam-cpp-sdk.yml
@@ -1,0 +1,121 @@
+name: Bump Viam C++ SDK
+
+# Opens (or updates) a PR bumping the viam-cpp-sdk version whenever a newer
+# viam-cpp-sdk release is published, polled daily. The SDK is pinned via conan
+# (which Dependabot can't track) AND duplicated as a git tag across several build
+# files, so this updates every location in one PR.
+#
+# SDK-version locations kept in sync (auto-discovered, so new ones are covered):
+#   - Dockerfile / bin/setup.sh / CMakeLists.txt : `releases/vX.Y.Z` git tag
+#   - conanfile.py                               : `viam-cpp-sdk/[>=X.Y.Z]` range
+#   - CMakeLists.txt (yaskawa only)              : `find_package(viam-cpp-sdk X.Y ...)`
+#
+# The PR is created with the org-level GIT_ACCESS_TOKEN bot secret (the same token
+# the viam-python-sdk update_protos workflow uses) rather than the default
+# GITHUB_TOKEN. This is deliberate: PRs opened by GITHUB_TOKEN do NOT trigger other
+# workflows, so the build/test wouldn't run on the bump; the bot token makes CI run.
+
+on:
+  schedule:
+    - cron: "0 6 * * *" # daily at 06:00 UTC
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  bump:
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-viam-cpp-sdk.yml
```diff
@@ -0,0 +1,121 @@
+name: Bump Viam C++ SDK
```
acmorrow, line 1: I'm not entirely sure I agree with this change. We have a ticket https://viam.atlassian.net/browse/RSDK-12643 that falls into the Yaskawa/UR/Trajex CI improvements epic to address exactly this issue: that the version is specified in too many places. By my count, four in Universal Robots.

But there are two different classes of version numbers in play here:

The version numbers to `FIND_PACKAGE_ARGS` and in `conanfile.py` are version _minima_. They state what we need either to not build the C++ SDK from source ourselves (i.e. accept an externally built version, from the system or elsewhere), or to demand from the conan environment when building.

The version numbers on the `GIT_TAG` in the `CMakeLists.txt` file and the `git checkout` in the setup script are the: what version are we going to build, when we need to do so.

The minima should only be updated when the universal robots module becomes dependent on a something in a newer version of the C++ SDK. Otherwise, someone with a perfectly good install of the C++ SDK can't use it. The "what will we build" versions should be updated whenever C++ SDK has a release that we can accept without other changes. But an update that introduces a breaking change needs a bigger PR.

We've historically done a poor job of distinguishing the two things going on here, and have often updated the versions in lockstep, as this PR does.

We also have major changes on the horizon to how we do conan stuff in general, with the work that @lia-viam is doing around docker containers and caching.

So, I think I would like us to first try to:
1) Do a better job distinguishing the two version number roles (min vs build)
2) Get on the other side of the conan work
3) Try to reduce the number of places we need a version number at all. Ideally there would just be two: the minimum in one place, and the build target in another.

And then maybe this PR can be much simpler?

acmorrow (changes requested):

.github/workflows/bump-viam-cpp-sdk.yml
stuqdog: That all makes sense! Sounds like probably the best thing to do is to close this and the yaskawa PRs then, and then we can revisit once the CI improvements work is done?

PR #222, opened 2026-07-09
telemetry: read the trace ID from a configurable metadata field

vijayvuyyuru:
The telemetry output path is tagged by the trace ID parsed out of the incoming `traceparent` header. Callers that have moved off OpenTelemetry to RDK's request-metadata propagation no longer set `traceparent`, so there's no way to tag telemetry with the identifier they do propagate.

Add an optional `traceid_metadata_key` config attribute. When set, the trace ID is read directly from that request-metadata field; when unset, it falls back to parsing the `traceparent` header as before. The existing `telemetry_output_path_append_traceid` / `{trace_id}` behavior is unchanged, keeping the module domain-agnostic — an operator configures the key per arm to whatever field carries their identifier.

## Testing:
Tested manually in lab. Saw matching files to main generated from a pass (****_arm_realtime_trajectory.json)

### DIFF
--- a/src/viam/ur/module/ur_arm_state.cpp
+++ b/src/viam/ur/module/ur_arm_state.cpp
@@ -44,6 +44,7 @@ URArm::state_::state_(private_,
                       std::optional<vector6d_t> max_acceleration_limits,
                       double trajectory_sampling_freq_hz,
                       std::string telemetry_output_path_append_traceid_template,
+                      std::string traceid_metadata_key,
                       const struct ports_& ports)
     : configured_model_{std::move(configured_model)},
       resource_name_{std::move(resource_name)},
@@ -65,7 +66,8 @@ URArm::state_::state_(private_,
       max_velocity_limits_(std::move(max_velocity_limits)),
       max_acceleration_limits_(std::move(max_acceleration_limits)),
       trajectory_sampling_freq_hz_(trajectory_sampling_freq_hz),
-      telemetry_output_path_append_traceid_template_(std::move(telemetry_output_path_append_traceid_template)) {
+      telemetry_output_path_append_traceid_template_(std::move(telemetry_output_path_append_traceid_template)),
+      traceid_metadata_key_(std::move(traceid_metadata_key)) {
     // Initialize the kinematics promise/future pair so that
     // `kinematics_future_.valid()` is unconditionally true for the rest of
     // `state_`'s lifetime. The producer side (see
@@ -164,6 +166,10 @@ std::unique_ptr<URArm::state_> URArm::state_::create(UrArmModel configured_model
         return {};
     }();
 
+    // Parse `traceid_metadata_key` — the request-metadata field the trace id is read from. When
+    // unset, the trace id is derived from the traceparent header (see telemetry_output_path).
+    const auto traceid_metadata_key = find_config_attribute<std::string>(config, "traceid_metadata_key").value_or("");
+
     // Look for the optional settings to place an upper bound on what velocity and acceleration limits may be
     // configured, or set via DoCommand or applied via MoveOptions. Note that the parse/validate function automatically
     // converts from degrees to radians.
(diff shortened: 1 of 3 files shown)


src/viam/ur/module/utils.hpp
```diff
@@ -134,24 +134,34 @@ urcl::vector6d_t parse_and_validate_joint_limits(const viam::sdk::ProtoValue& va
 urcl::vector6d_t parse_and_validate_joint_limits(const viam::sdk::ResourceConfig& cfg, const std::string& param_name);
 
 ///
-/// Extract trace-id from W3C Trace Context traceparent header.
+/// gRPC metadata header carrying the sanding pass ID. RDK prefixes arbitrary
+/// request metadata keys with "viam-metadata-" on the wire, so a value set as
+/// "sanding-pass-id" is received as "viam-metadata-sanding-pass-id".
+///
+inline constexpr std::string_view k_pass_id_metadata_key = "viam-metadata-sanding-pass-id";
```
acmorrow: Hey, sorry, I guess I should have inquired further when we chatted about this. I don't think it makes sense for the universal robots module to be aware of sanding specific metadata fields. I think if we want to do this, then this needs to become configurable, under something like a new config key that describes what metadata field to use for tracing. Then you could configure it for a specific arm to be `viam-metadata-sanding-pass-id` and that would be fine.

Similarly, I think `pass` is a sanding specific term.

Could we achieve all of this by keeping the existing `telemetry_output_path_append_traceid` with its current semantics, and then introducing a new `traceid_metadata_key` or similar? If not set, it will use the existing traceparent stuff, as the key (just make it the default) but if it is set, it just looks for that field in the metadata.

acmorrow (changes requested):

src/viam/ur/module/utils.hpp
vijayvuyyuru: Hey sorry, I didnt mean to tag you on this yet, should have left as a draft. I agree with what youre saying here.

acmorrow: No problem, just lmk when you are ready for me to have a look

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -44,6 +44,7 @@ URArm::state_::state_(private_,
                       std::optional<vector6d_t> max_acceleration_limits,
                       double trajectory_sampling_freq_hz,
                       std::string telemetry_output_path_append_traceid_template,
+                      std::string traceid_metadata_key,
```
acmorrow, line 47: Not your doing, but seeing this here reminds me that we need to do something about this constructor. That is just far too many args.

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -164,6 +166,10 @@ std::unique_ptr<URArm::state_> URArm::state_::create(UrArmModel configured_model
         return {};
     }();
 
+    // Parse `traceid_metadata_key` — the request-metadata field the trace id is read from. When
+    // unset, the trace id is derived from the traceparent header (see telemetry_output_path).
+    const auto traceid_metadata_key = find_config_attribute<std::string>(config, "traceid_metadata_key").value_or("");
```
acmorrow, line 171: Would it simplify things if this was `.value_or("traceparent")`?

src/viam/ur/module/ur_arm_state.cpp
```diff
@@ -315,15 +322,22 @@ std::filesystem::path URArm::state_::telemetry_output_path() const {
...
+    std::optional<std::string> trace_id;
+    if (!traceid_metadata_key_.empty()) {
+        const auto values = observer->get_client_metadata_field_values(traceid_metadata_key_);
+        if (!values.empty()) {
+            trace_id = values[0];
+        }
+    } else {
+        const auto traceparent_values = observer->get_client_metadata_field_values("traceparent");
+        if (!traceparent_values.empty()) {
+            trace_id = extract_trace_id_from_traceparent(traceparent_values[0]);
```
acmorrow, line 336: I think that answers my question above: it wouldn't simplify things to default the parameter, because then you would lose the signal for whether you needed to call `extract_trace_id...` on the result or not.

acmorrow (approved):
LGTM

PR #223, opened 2026-07-16
[RSDK-14161] Implement support for streaming

acmorrow:


src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -96,12 +98,60 @@ class URArm::state_ {
     double get_trajectory_sampling_freq_hz() const;
 
     void clear_pstop() const;
-    void zero_ftsensor() const;
-
-    size_t get_move_epoch() const;
 
+    ///
```
acmorrow: These shouldn't be doxygen comments, will change.

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -398,6 +445,46 @@ class URArm::state_ {
         std::string_view describe() const;
     };
 
+    // Shared between the producer and the worker to track an open-ended
+    // trajectory stream. A default-constructed one is freshly opened: k_open,
```
acmorrow: comments are inconsistent about whether formal names are in `ticks` or not, will ask claude to fix.

src/viam/ur/module/trajectory_logger.cpp
```diff
@@ -1,22 +1,53 @@
...
-    root_["timestamp"] = timestamp;
+    const auto move_id_str = boost::uuids::to_string(move_id);
+    root_["move_id"] = move_id_str;
+    root_["timestamp"] = make_iso8601_now();
     root_["robot_model"] = robot_model;
     root_["resource_name"] = resource_name;
     root_["realtime_samples"] = Json::Value(Json::arrayValue);
 
-    output_path_ = telemetry_path / (timestamp + "_" + resource_name + "_realtime_trajectory.json");
+    output_path_ = telemetry_path / (move_id_str + "_" + resource_name + "_realtime_trajectory.json");
```
JohnN193, line 50: will this make it harder to pair moves with sanding data?

src/viam/ur/module/ur_arm_state_controlled.cpp
```diff
@@ -67,57 +67,73 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::h
...
-                        return event_connection_lost_::trajectory_control_failure();
+            if constexpr (std::is_same_v<T, sample_stream>) {
+                // Joint-space streaming. The phase drives what we send to URCL;
+                // see `sample_stream` for what each phase means. The phase only
+                // moves forward:
+                //   from k_open, first points pending: send STREAM_START, go to k_streaming.
+                //   from k_open, on close: go to k_buffered.
+                //   from k_streaming, on close: go to k_draining.
+                //   from k_buffered: send START, drain, and END in one tick, go to k_ended.
+                //   from k_draining, once drained: send STREAM_END, go to k_ended.
```
JohnN193: having trouble finding it, but just to confirm does `move_through_joint_positions `just go through the k_buffered branch? or do we call `TRAJECTORY_START` somewhere else for that now?

JohnN193 (commented):
not really sure how users/sanding will use the uuid. streaming stuff looks good to me though

src/viam/ur/module/ur_arm_state_controlled.cpp
acmorrow: That's exactly right: `move_through_joint_positions` will call `start_move_request` with `trajectory_samples`, which will find the matching `move_request` constructor. That constructor will create a `sample_stream` that begins life in `k_buffered`, with all the points already there. Then the state machine knows that for `k_buffered`, there is no need to use the `STREAM` stuff and falls back to plain old `TRAJECTORY_START`. A little convoluted, but has the net effect that both kinds of moves arrive in the state machine the same way, and we only stream when streaming is needed. A one-shot move becomes k_buffered and goes the old way, as does a, say, degenerate stream with only one batch of points that all arrive before the worker thread tick.

src/viam/ur/module/trajectory_logger.cpp
acmorrow: It shouldn't. Before, the timestamp was being used as the unique id, basically. It got tacked onto files and logs, but it was weird, because you would see the logs advancing with real time, but the logged "unix time" was the "correlation id". Now you just get a UUID for each move: way easier to correlate, and isn't confusing in the logs where there are two "times".

That said, we should definitely alert them to the change!

acmorrow: I checked in with @bgrigsby8 and he was OK with this change.

JohnN193 (approved):

PR #224, opened 2026-07-24
[RSDK-14161] Upgrade dockerfile to new C++ SDK

acmorrow:
Trivial, but a pre-req to upgrading the requirements in our CMakeLists and the tags we pull for CI lint/test runs.


JohnN193 (approved):
Should I do the same for yaskawa?

acmorrow:
> Should I do the same for yaskawa?

 No rush on that one since streaming isn't implemented for it.

PR #225, opened 2026-08-13
[RSDK-14378] Fix SDK typo and forward point counts correctly

acmorrow:


src/viam/ur/module/ur_arm_state_controlled.cpp
```diff
@@ -135,8 +135,9 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::h
                 }
                 if (cancel_pending && !cancel_issued) {
                     state.move_request_->cancellation_request->issued = true;
-                    if (!arm_conn_->driver->writeTrajectoryControlMessage(
-                            urcl::control::TrajectoryControlMessage::TRAJECTORY_CANCEL, 0, RobotReceiveTimeout::off())) {
+                    if (!arm_conn_->driver->writeTrajectoryControlMessage(urcl::control::TrajectoryControlMessage::TRAJECTORY_CANCEL,
+                                                                          static_cast<int>(cmd.points_written),
```
acmorrow, line 139: Important part here is sending the total point count when canceling: that's a requirement in the new URCL streaming support, and we were failing to honor it.

Note that URCL has its own issues (see https://github.com/UniversalRobots/Universal_Robots_Client_Library/issues/550), but this definitely wasn't helping.

src/viam/ur/module/ur_arm_state_controlled.cpp
```diff
@@ -285,7 +286,7 @@ std::optional<URArm::state_::event_variant_> URArm::state_::state_controlled_::h
                     // and cancellation is requested but has not yet been issued. Issue a cancel.
                     state.move_request_->cancellation_request->issued = true;
                     if (!arm_conn_->driver->writeTrajectoryControlMessage(
-                            urcl::control::TrajectoryControlMessage::TRAJECTORY_CANCEL, 0, RobotReceiveTimeout::off())) {
+                            urcl::control::TrajectoryControlMessage::TRAJECTORY_CANCEL, 1, RobotReceiveTimeout::off())) {
```
acmorrow, line 289: Technically, this isn't required, but it is harmless to provide an accurate count on cancel, so rather than trying to have this code know when to do it and when not, just always do it.

CMakeLists.txt
```diff
@@ -135,7 +135,7 @@ FetchContent_MakeAvailable(Universal_Robots_Client_Library)
 FetchContent_Declare(
   viam-cpp-sdk
   GIT_REPOSITORY https://github.com/viamrobotics/viam-cpp-sdk
-  GIT_TAG v0.39.0
+  GIT_TAG releases/v0.39.0
```
acmorrow, line 138: This didn't cause problems in CI because in the docker container, we have C++ SDK pre-instaleld so we never try to fetch the code. It failed immediately though anywhere without the SDK pre-installed.

JohnN193 (approved):

PR #227, opened 2026-08-19
[RSDK-14385] Emit per-joint velocity and acceleration limits

JohnN193:
Ticket 7 of [RSDK-14385](https://viam.atlassian.net/browse/RSDK-14385). The SVA kinematics schema now carries optional per-joint `max_velocity` and `max_acceleration` so the motion service can generate trajectories itself instead of handing arms a path and asking them to build a kinematic chain. This publishes them from our synthesized document.

## What changed

- `JointLimits` gains `max_velocity_deg_per_sec` and `max_acceleration_deg_per_sec2` as `std::optional`, because an absent field is how the schema spells unbounded and the shipped `kinematics/<model>.json` files carry position bounds only.
- `Kinematics::with_kinematic_limits` attaches per-joint limits, taking radians (what `state_` holds) and converting to the degrees the schema uses.
- `to_sva_json` writes each field only when set. `load_kinematics` reads them back, so a document we emit reparses to what we wrote.
- `state_::create` captures the configured limits once, and `get_dh_kinematics_json` stamps them inside the existing `call_once`.

## What we publish, and one deliberate tradeoff

The values are the configured `speed_degs_per_sec` and `acceleration_degs_per_sec2`, already clamped against any `max_speed_degs_per_sec` / `max_acceleration_degs_per_sec2` ceiling, so what we advertise is what an un-overridden move will use. The ceilings are never the effective value: `clamp_velocity_limits` applies them on the way in, so a configured 30 under a ceiling of 100 runs at 30, and publishing 100 would be a looser bound than the arm honors.

We snapshot at configure time rather than reading the live limits. **`set_speed_degs_per_sec` via DoCommand therefore does not change what `GetKinematics` reports.** That is intentional, not an oversight:

- The framesystem reads kinematics once when it builds the resource graph and never re-reads it, so live values would not reach the planner anyway.
- Publishing live values would only make RDK's two read paths disagree with each other, since `armplanning.MoveArm` fetches per call while the framesystem does not.
- It would also make the returned bytes change under anything hashing them.
- Publishing the configured value matches the yaskawa module, so the motion service gets the same meaning from both arms.

A consequence worth knowing: after `DoCommand{"set_speed_degs_per_sec": 5}` on an arm configured at 60, the document still reports 60 while MoveThrough/ToJointPositions plan against 5.

## Testing

- `test_to_sva_json_round_trips_via_parse` extended to stamp per-joint-distinct limits (110-160 deg/s, 210-260 deg/s²) so a transposed or broadcast write fails, and to check position bounds survive.
- `test_configured_zero_limits_are_published_as_zero` pins the zero behavior in both the parsed optionals and the emitted JSON.
- `test_unstamped_kinematic_limits_are_omitted` checks an unstamped document has neither key.

`make test` green, `-Wall -Werror` clean, clang-tidy and clang-format clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[RSDK-14385]: https://viam.atlassian.net/browse/RSDK-14385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/ur/module/ur_arm_model.cpp
+++ b/src/viam/ur/module/ur_arm_model.cpp
@@ -6,6 +6,7 @@
 #include <fstream>
 #include <initializer_list>
 #include <memory>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
@@ -253,7 +254,22 @@ JointLimits parse_joint_limits(const std::filesystem::path& path, const Json::Va
     if (!joint.isMember("max") || !joint["max"].isNumeric()) {
         throw_parse_error(path, "joint missing numeric `max`");
     }
-    return JointLimits{joint["min"].asDouble(), joint["max"].asDouble()};
+
+    // `max_velocity` and `max_acceleration` are optional in the schema, so an
+    // absent field is not an error, but a present one that is not a number is.
+    // We read them so a document we emitted parses back to what we wrote.
+    const auto optional_limit = [&](const char* field) -> std::optional<double> {
+        if (!joint.isMember(field)) {
+            return std::nullopt;
+        }
+        if (!joint[field].isNumeric()) {
+            throw_parse_error(path, std::string{"joint `"} + field + "` is present but not a number");
+        }
+        return joint[field].asDouble();
+    };
+
+    return JointLimits{
+        joint["min"].asDouble(), joint["max"].asDouble(), optional_limit("max_velocity"), optional_limit("max_acceleration")};
 }
 
 Json::Value translation_json(const Eigen::Vector3d& t) {
@@ -396,6 +412,19 @@ UrArmModel::Kinematics UrArmModel::Kinematics::apply_calibration(const DHParams&
     }
 
     out.link_locals = new_link_locals;
+    return out;
+}
+
+UrArmModel::Kinematics UrArmModel::Kinematics::with_kinematic_limits(const urcl::vector6d_t& velocity_rad_per_sec,
+                                                                     const urcl::vector6d_t& acceleration_rad_per_sec2) const {
+    // We publish whatever is configured, zero included. Only an absent field means unbounded, so a
+    // configured zero has to go out as zero: our validation accepts an array with a single zero
+    // element, and omitting it would tell the planner that an axis which cannot move is unbounded.
+    Kinematics out = *this;
+    for (std::size_t i = 0; i < k_num_dh_joints; ++i) {
+        out.limits[i].max_velocity_deg_per_sec = radians_to_degrees(velocity_rad_per_sec[i]);
+        out.limits[i].max_acceleration_deg_per_sec2 = radians_to_degrees(acceleration_rad_per_sec2[i]);
+    }
     return out;
 }
 
@@ -566,6 +595,14 @@ std::string UrArmModel::Kinematics::to_sva_json() const {
             joint["axis"] = dh_z_axis;
             joint["min"] = limits[i - 1].min_deg;
             joint["max"] = limits[i - 1].max_deg;
+            // Leaving these out is how the schema spells "unbounded", so we
+            // emit nothing rather than a placeholder when we have no value.
+            if (limits[i - 1].max_velocity_deg_per_sec) {
+                joint["max_velocity"] = *limits[i - 1].max_velocity_deg_per_sec;
+            }
+            if (limits[i - 1].max_acceleration_deg_per_sec2) {
+                joint["max_acceleration"] = *limits[i - 1].max_acceleration_deg_per_sec2;
+            }
             joints.append(joint);
         }
 
--- a/src/viam/ur/module/ur_arm_model.hpp
+++ b/src/viam/ur/module/ur_arm_model.hpp
@@ -43,11 +43,18 @@ struct DHParams {
     urcl::vector6d_t theta;
 };
 
-// Joint angular limits, in degrees, matching the `min`/`max` fields in
-// shipped `kinematics/<model>.json` files.
+// Joint limits, in degrees, matching the `min`/`max`/`max_velocity`/
+// `max_acceleration` fields in shipped `kinematics/<model>.json` files.
+//
+// The two kinematic limits are optional because the SVA schema treats an
+// omitted field as unbounded, and because the shipped files carry position
+// bounds only. They are filled in per arm instance from configured speed and
+// acceleration by `Kinematics::with_kinematic_limits`.
 struct JointLimits {
     double min_deg;
     double max_deg;
+    std::optional<double> max_velocity_deg_per_sec;
+    std::optional<double> max_acceleration_deg_per_sec2;
 };
 
 // A geometry expressed in its link's parent (joint) frame.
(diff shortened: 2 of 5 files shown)


src/viam/ur/module/ur_arm_model.hpp
```diff
@@ -125,6 +132,11 @@ class UrArmModel::Kinematics {
 
     Kinematics apply_calibration(const DHParams& dh) const;
 
+    // Returns a copy carrying the given per-joint velocity and acceleration
+    // limits. Inputs are radians, matching what `URArm::state_` holds; the
+    // conversion to the degrees the SVA schema uses happens here.
+    Kinematics with_kinematic_limits(const urcl::vector6d_t& velocity_rad_per_sec, const urcl::vector6d_t& acceleration_rad_per_sec2) const;
```
acmorrow: Above, these builder style methods are called `apply_x`, and I recommend continuing that pattern here: `apply_kinematic_limits`.

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -399,6 +415,19 @@ UrArmModel::Kinematics UrArmModel::Kinematics::apply_calibration(const DHParams&
...
+UrArmModel::Kinematics UrArmModel::Kinematics::with_kinematic_limits(const urcl::vector6d_t& velocity_rad_per_sec,
+                                                                     const urcl::vector6d_t& acceleration_rad_per_sec2) const {
+    // We publish whatever is configured, zero included. Only an absent field means unbounded, so a
+    // configured zero has to go out as zero: our validation accepts an array with a single zero
+    // element, and omitting it would tell the planner that an axis which cannot move is unbounded.
+    Kinematics out = *this;
+    for (std::size_t i = 0; i < k_num_dh_joints; ++i) {
+        out.limits[i].max_velocity_deg_per_sec = radians_to_degrees(velocity_rad_per_sec[i]);
+        out.limits[i].max_acceleration_deg_per_sec2 = radians_to_degrees(acceleration_rad_per_sec2[i]);
+    }
```
acmorrow, line 435: Presumably, if we have a value for one of these limits from the SVA, that's a true hardware constraint. Should it be possible to increase a limit here? Or should these be capped at what the SVA says the limit really is.

In other words, should we only allow the runtime alteration based on configuration to be a limit down move? Should it be an error to try to increase beyond what the SVA says?

And, what about limits as configured on the robot itself? I believe UR, for instance, offers on-device safety limits on position and velocity. How should those be handled? Should those be involved in our synthesized return?

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -566,6 +595,14 @@ std::string UrArmModel::Kinematics::to_sva_json() const {
             joint["axis"] = dh_z_axis;
             joint["min"] = limits[i - 1].min_deg;
             joint["max"] = limits[i - 1].max_deg;
+            // Leaving these out is how the schema spells "unbounded", so we
```
acmorrow: What would such an unbounded velocity or acceleration mean, physically? Should we accept an SVA that constrains only a subset of the joints? I feel like either you know the limits, or you don't.

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -594,6 +602,23 @@ class URArm::state_ {
...
 
+    // The configured limits, already clamped against any ceiling, captured once during `create`
+    // and never touched again. The kinematics document describes how this arm is configured, so it
+    // publishes these rather than the live `velocity_limits_`.
+    //
+    // That is a deliberate tradeoff, not just a convenience. `set_speed_degs_per_sec` via DoCommand
+    // moves the live limits for the rest of the session, and moves plan against those, so after
+    // such a call the document reports a speed the arm will not use. We accept that because the
+    // framesystem reads kinematics once when it builds the resource graph and never re-reads it, so
+    // publishing live values would not reach the planner anyway; it would only make the two RDK
```
acmorrow: We are going to need, in the more complete kinematics project, a means to address caching. Consider, for instance, that an arm could be connected to physical arm 1 at a particular IP address, lose connection because that robot arm was taken out of service, and then regain connection to physical arm 2 at that same IP address because it is rotated into service to replace arm 1. The calibration on arm 2 could be different. Without a mechanism to have the arm module signal that kinematics may have changed, the client will be using the wrong one. If you are tracking work items for the next scope, I think it is important to get kinematics cache invalidation on the list.

acmorrow (commented):
I'm going to focus on this one first rather than the Yaskawa one, because I know this codebase better. Mostly questions at this point. It is a little tricky to review because I keep running into the limits of the current approach and seeing the problems, rather than seeing the gains we make by adopting what is here, despite the rough edges.

src/viam/ur/module/ur_arm_model.cpp
JohnN193: I had intended to only consider the configured values for vel/acc to begin with, rather than having to update all the SVAs. I would rather not modify the actual SVAs just yet, but if its preferred I can try to find the numbers.

I think the short answer is yes, but I would rather save that kind of work for a future kinematics project. When talking with Dan we both felt that whatever we send over the wire should contain both the robot limit and the user's configured limit. That way if another application wants to modify the vel/acc limits they have all of the information to do so. 

Def think at some point we should read from the limits configured on the robot to validate against, since we probably should already be doing that. I intended to make followup tickets for both xarm and ur to do so

src/viam/ur/module/ur_arm_model.cpp
JohnN193: unbounded would mean that your planner should try to move the arm as fast as possible, hopefully with some other constraint in mind. leaving it unbounded could also work for a simulated environment. 

In practice I needed some way to describe an optional state where trajex shouldn't be used for the motion service to grab, as we continue this kinematics work. Basically so arms that still using URDFs or aren't sending kinematic limits won't break in the motion service. the check I added to [rdk checks if any of the joints are unset](https://github.com/viamrobotics/rdk/blob/c2c9600c60bb0fba2ce2648a47a6a3b71ed96e38/referenceframe/frame.go#L170-L183). 

For this pr its both impossible to actually send unbounded joints, and in the ur module where the hardware has opinions on how fast motors can move I will go ahead and remove any sort of "unbounded" possibilities.

src/viam/ur/module/ur_arm_model.cpp
```diff
@@ -253,7 +254,30 @@ JointLimits parse_joint_limits(const std::filesystem::path& path, const Json::Va
...
+    // as a negative speed limit. Zero is allowed, since that is a real limit meaning the joint does
+    // not move. Unlike `min` and `max`, which are positions and are routinely negative.
+    const auto optional_limit = [&](const char* field) -> std::optional<double> {
+        if (!joint.isMember(field)) {
+            return std::nullopt;
+        }
+        if (!joint[field].isNumeric()) {
+            throw_parse_error(path, std::string{"joint `"} + field + "` is present but not a number");
+        }
+        const double value = joint[field].asDouble();
```
acmorrow: FYI This is an ideal case for `const auto value`, since the `asDouble` makes extra clear what type you will be getting.

src/viam/ur/module/ur_arm_state.hpp
```diff
@@ -594,6 +598,20 @@ class URArm::state_ {
     vector6d_t velocity_limits_{};
     vector6d_t acceleration_limits_{};
 
+    // The configured limits, already clamped against any ceiling, captured once during `create`.
+    // The kinematics document describes how this arm is configured, so it publishes these rather
+    // than the live `velocity_limits_`, which DoCommand moves for the session and MoveOptions moves
+    // for a single move. That means the document is not an upper bound on what the arm will do; see
```
acmorrow: I think I'd prefer the justifcation be inlined here, rather than separate in the PR. Or, if the justification is quite lengthy, I could live with a link to (this?) PR.

acmorrow (approved):
LGTM.

